PR #6, opened 2021-06-16
ISAICP-6549: Deploy the codebase in only one step

claudiu-cristea:
Jira issue(s):
- [ISAICP-6551](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-6551)

The OE Bootstrap Theme is Drupal theme, hence is designated for PHP developers. Such developers are more familiar with Composer rather than NPM.

We propose that we add `npm install` as part of `oe_bootstrap_theme:npm-run` command and we run it only if there's no `package-lock.json` present. This allows to build the codebase with just one command: `composer install`

### DIFF
--- a/src/TaskRunner/Commands/BuildCommands.php
+++ b/src/TaskRunner/Commands/BuildCommands.php
@@ -46,6 +46,10 @@ class BuildCommands extends AbstractCommands {
       ->dir($npmDir)
       ->stopOnFail();
 
+    if (!file_exists($npmDir . '/package-lock.json')) {
+      $stack->exec('npm install');
+    }
+
     // If node_modules dir doesn't exist, run 'npm ci'.
     // See https://stackoverflow.com/questions/52499617/what-is-the-difference-between-npm-install-and-npm-ci
     if (!is_dir($npmDir . '/node_modules')) {
--- a/.drone.yml
+++ b/.drone.yml
@@ -13,13 +13,6 @@ services:
       - MYSQL_ALLOW_EMPTY_PASSWORD=yes
 
 pipeline:
-  npm-build:
-    group: prepare
-    image: node:9.5.0
-    commands:
-      - npm install --unsafe-perm
-      - NODE_ENV=production npm run production
-
   composer-install:
     group: prepare
     image: fpfis/httpd-php-ci:${PHP_VERSION}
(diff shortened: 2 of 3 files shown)


src/TaskRunner/Commands/BuildCommands.php
```diff
@@ -46,6 +46,10 @@ public function npmRun(string $npmCommand, array $options = [
       ->dir($npmDir)
       ->stopOnFail();
 
+    if (!file_exists($npmDir . '/package-lock.json')) {
```
brummbar, line 49: What is the scenario on this? The file gets removed if a new version of the theme is downloaded? Cause dependencies might be updated, so the lock is outdated at that moment.

brummbar (changes requested):
This is not necessary, as developers that are using this theme as dependency should use a release package and not the "normal" tagged version. When a release is tagged, a package will be built. By using the composer artefacts composer plugin, you can fetch the fully built package, with the compiled assets and all, without needing to use npm in production, or even in local environments where you are not developing on the theme.

src/TaskRunner/Commands/BuildCommands.php
pfrenssen: I think we should probably commit the `package-lock.json` file and update dependencies periodically. This makes it possible for the theme developers to install a working version of the development environment, also for historic versions.

I am struggling right now with a website that doesn't commit the lock file. The theme is now 2 years old and it is practically impossible to figure out which versions I need to install to make the gulp commands work. I get recent versions of the dependencies and now the gulp commands are throwing errors, and now I need to do detective work in changelogs to find out which exact versions I need for these commands to work. If the lock file was committed, then I would get the exact versions and I would be able to compile the SASS and minify assets.

I think the main idea is:
* Production code of libraries / modules / plugins: do not commit lock file, the composer / package files should declare _the range of versions the library is compatible with_.
* Development code of libraries (and production code of actual projects): commit lock files.

So in the case of the OE Bootstrap theme: we should not commit `composer.lock` (= production code intended to be compatible with a range of Drupal versions) but we should commit `package-lock.json` (= a dev environment in which the scripts currently work).

It is also necessary for replicatable builds to have a record of the exact versions of the SASS compiler etc that were used to build the assets. With other versions the generated code might be different and the build is no longer replicatable.

PR #38, opened 2021-07-26
ISAICP-6548: Preserve link attributes in menus

pfrenssen:
When using the theme in Joinup we are seeing test failures in the navigation menus.

After investigation it appears that the menu links are rendered by creating the `<a>` tag from scratch, preserving only the `class` and `target` attributes. All other attributes that might be present on the link are ignored. We are missing things such as the `title` attribute, aria attributes, and Drupal `data-` attributes which are needed for JavaScript behaviours.

A good solution would be to use the `link()` Twig function which preserves all attributes.


escuriola:
I think this is not happened any more. Can I close @brummbar @drishu?

PR #45, opened 2021-08-12
FRONT-2449: Button Pattern.

### LINKED ISSUE #16: Refactor Button pattern with BCL component

tibi2303:


drishu (changes requested):
see todo in templates/overrides/form/input--submit.html.twig
this needs to be implemented.
Take notice of button attributes (primary secondary) and change according to the pattern variants.

Please add `- "bcl"` under `file_scan_ignore_directories` in runner.dist.yml until https://github.com/openeuropa/oe_bootstrap_theme/issues/61 is solved.

escuriola:
> see todo in templates/overrides/form/input--submit.html.twig
> this needs to be implemented.
> Take notice of button attributes (primary secondary) and change according to the pattern variants.
> 
> Please add `- "bcl"` under `file_scan_ignore_directories` in runner.dist.yml until #61 is solved.

This would print a button which by itself that not submit the form.

drishu:
> > see todo in templates/overrides/form/input--submit.html.twig
> > this needs to be implemented.
> > Take notice of button attributes (primary secondary) and change according to the pattern variants.
> > Please add `- "bcl"` under `file_scan_ignore_directories` in runner.dist.yml until #61 is solved.
> 
> This would print a button which by itself that not submit the form.

I think that if the button is of type submit then its fine, please test

drishu (changes requested):
I think for this task we should theme button elements also, I looked into it, to do this we can't override a template, but we can theme from scss
in _all.scss add
```
button.link {
  display: inline;
  margin: 0;
  padding: 0;
  cursor: pointer;
  border: 0;
  background: transparent;
  font-size: 1em;
  text-decoration: underline;
  appearance: none;
}
```
this is needed so that the "Edit summary" button looks ok in create content form (and other forms )

templates/patterns/button/pattern-button.html.twig
```diff
@@ -1,18 +1,28 @@
...
+  'type': type,
+  'outline': outline,
+  'disabled': disabled,
+  'text_nowrap': text_nowrap,
+  'assistive_text': assistive_text,
+  'id': id,
+  'icon': _icon,
+  'icon_position': icon_position,
+  'clean_class': clean_class,
+  'spinner': spinner,
```
drishu, line 34: what is this, I don't see it in the field list

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -26,20 +26,56 @@ button:
...
+    icon_position:
+      type: select
+      label: Icon position
+      options:
+        before: before
+        after: after
+    show_spinner:
+      type: boolean
+      label: Show spinner
+    id:
```
drishu: this is a reserver word, use soemthing else

drishu (changes requested):

drishu (changes requested):
Looks good, but one more hurdle to pass, the functionality of other patterns was affected because the fields and options changed, please fix:
pattern-collapse.html.twig
pattern-dropdown.html.twig
pattern-modal.html.twig
pattern-popover.html.twig

Maxfire:
> Looks good, but one more hurdle to pass, the functionality of other patterns was affected because the fields and options changed, please fix:
> pattern-collapse.html.twig
> pattern-dropdown.html.twig
> pattern-modal.html.twig
> pattern-popover.html.twig

I checked all and collapse was fixed, also checked different variants, greetings.

templates/patterns/button/pattern-button.html.twig
```diff
@@ -1,18 +1,34 @@
...
+{#
+/**
+ * @file
+ * Button component.
+ */
+#}
+{%  if _icon %}
+  {% set _icon = icon|merge({
+    path: bcl_icon_path
+})  %}
```
drishu: intendation is not correct, and there is a extra space
```suggestion
  }) %}
```

templates/patterns/button/pattern-button.html.twig
```diff
@@ -1,18 +1,34 @@
...
-  <a{{ attributes.setAttribute('href', url) }}>{{ label }}</a>
-{% else %}
-  {% if disabled %}
-    {% set attributes = attributes.setAttribute('disabled', true) %}
-  {% endif %}
-  <button type="button"{{ attributes }}>{{ label }}</button>
+{#
+/**
+ * @file
+ * Button component.
```
drishu: * Button pattern.

templates/overrides/form/input--submit.html.twig
```diff
@@ -10,5 +10,13 @@
  * @see template_preprocess_input()
  */
 #}
-{# @TODO Revisit after refactoring the button pattern. #}
-<input{{ attributes.addClass(['btn', 'btn-primary']) }} />{{ children }}
+{% if attributes.hasClass('button--primary') %}
```
drishu: in the login form, the class `button--primary` is not set for the login button, it seems that this class is added per use case and not automatically by a rule.
To solve this issue, we can do something like ``{% elseif attributes['data-drupal-selector'] == 'edit-submit' %}``

drishu (changes requested):

templates/patterns/button/pattern-button.html.twig
```diff
@@ -1,18 +1,34 @@
...
+ */
+#}
+{%  if _icon %}
```
drishu:
```suggestion
{%  if icon %}
```

templates/patterns/button/pattern-button.html.twig
```diff
@@ -1,18 +1,34 @@
...
+  {% set _icon = icon|merge({
+    path: bcl_icon_path
+})  %}
 {% endif %}
+{%  if show_spinner %}
+  {% set spinner = {
+    'type': "border",
+    'variant': "info",
+    'small': true,
```
drishu, line 17: the icon is to close to the text, please add ``attributes: create_attribute().addClass('me-1'),`` like in netlify data tab

templates/patterns/button/button.ui_patterns.yml
```diff
@@ -26,20 +35,60 @@ button:
...
       type: text
       label: Label
       description: The button label
       preview: Submit
+    assistive_text:
+      type: text
+      label: Assistive text
+      description: Additional text hidden with the .visually-hidden class.
+      preview: ''
     url:
```
drishu: url doesn't exist

drishu (changes requested):

templates/patterns/collapse/pattern-collapse.html.twig
```diff
@@ -18,13 +18,13 @@
       label: label,
       style: style,
```
drishu:
```suggestion
      variant: style,
```

drishu (changes requested):

templates/overrides/form/input--submit.html.twig
Maxfire: Ok I'm going to add it as or condition in the same if. thanks.

templates/patterns/collapse/pattern-collapse.html.twig
Maxfire: this could be a bit confusing but ok.

abel-santos-corral (approved):

Maxfire (approved):

drishu (approved):

PR #50, opened 2021-08-24
FRONT-2475: Dropdown pattern.

### LINKED ISSUE #20: Refactor Dropdown pattern with BCL component

escuriola:


abel-santos-corral (dismissed):

templates/patterns/dropdown/pattern-dropdown.html.twig
```diff
@@ -1,29 +1,17 @@
-{% set attributes = attributes.addClass([
-  'dropdown',
-  direction ? direction
-  ]) %}
+{#
+/**
+ * @file
+ * Dropdown component.
```
drishu:
```suggestion
 * Dropdown pattern.
```

drishu (changes requested):
Small thing (see below) and rebase pls.
also look into templates/overrides/navigation/links--dropbutton.html.twig, maybe we need to adapt ? I'm not imposing though, up to you.

templates/patterns/dropdown/pattern-dropdown.html.twig
escuriola: I tried to adapt but I think is not possible. Main reasons: 

- The button for instance sometimes is outlined but at dropdown we cannot set this variant, 
- Links are rendered link objects, so is not possible to pass as variable

drishu (approved):

abel-santos-corral (approved):

PR #53, opened 2021-08-25
FRONT-2478: List group pattern.

### LINKED ISSUE #22: Refactor List-group pattern with BCL component

escuriola:


GilNovacomm (commented):
While the pattern works as designed a few points should be reviewed:
The border around the component is missing.
Passing items to the pattern itself does not work so a pattern override template for this component needs to be implemented to preprocess the necessary variables.
There is no **List group** paragraph, **Listing item block** is an existing paragraph if the later is the one this pattern is supposed to theme then it's name it not obvious and should be corrected.

templates/patterns/list_group/list_group.ui_patterns.yml
```diff
@@ -1,39 +1,46 @@
...
+        secondary: ordered
+        success: actionable
+      preview: unordered
     flush:
+      type: boolean
       label: Flush
-      description: Add .list-group-flush to remove some borders and rounded corners to render list group items edge-to-edge in a parent container (e.g., cards).
+      description: Add flush to remove some borders and rounded corners to render list group items edge-to-edge in a parent container (e.g., cards).
     horizontal:
+      type: boolean
```
drishu: must be select and list the options

drishu (changes requested):

drishu (approved):

Maxfire (approved):
approved. :+1:

PR #54, opened 2021-08-25
FRONT-2479: Modal pattern.

### LINKED ISSUE #23: Refactor Modal pattern with BCL component

escuriola:


abel-santos-corral:
Handled at https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-416

templates/patterns/modal/pattern-modal.html.twig
```diff
@@ -1,15 +1,13 @@
...
-{% endif %}
+{#
+/**
+ * @file
+ * Modal component.
+ */
+#}
 
 {% block trigger %}
   {{ pattern('button', {
```
drishu, line 12: what if we want to use this like we do now here https://github.com/openeuropa/oe_whitelabel/blob/1.x/templates/overrides/navigation/links--language-block.html.twig#L26 where the trigger should be a link, also we may want to use an existing link and have the modal printer somewhere else, so this is what I suggest:
lets wrap the button pattern in an if, ex `if btn_label is not empty` then the trigger can come from soemwhere else and we dont care what it is

templates/patterns/modal/modal.ui_patterns.yml
```diff
@@ -1,81 +1,68 @@
...
-      type: boolean
-      label: Scrollable
-      description: When modals become too long for the user’s viewport or device, they scroll independent of the page itself.
-      preview: true
+        sm: "sm"
+        lg: "lg"
+        xl: "xl"
+        fullscreen: "fullscreen"
+      preview: lg
+    id:
```
drishu: id is a reserver work for the pattern, if I set this it breaks, use `modal_id` or something like that

drishu (changes requested):

drishu (approved):

abel-santos-corral (approved):

PR #78, opened 2021-09-24
Front 2490: Table pattern

Maxfire:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-426

### DIFF
--- a/templates/overrides/dataset/table.html.twig
+++ b/templates/overrides/dataset/table.html.twig
@@ -5,79 +5,9 @@
  *
  */
 #}
-
-{% set classes = ['table', 'table-hover'] %}
-<div class="table-responsive">
-  <table{{ attributes.addClass(classes) }}>
-    {% if caption %}
-      <caption>{{ caption }}</caption>
-    {% endif %}
-
-    {% for colgroup in colgroups %}
-      {% if colgroup.cols %}
-        <colgroup{{ colgroup.attributes }}>
-          {% for col in colgroup.cols %}
-            <col{{ col.attributes }} />
-          {% endfor %}
-        </colgroup>
-      {% else %}
-        <colgroup{{ colgroup.attributes }} />
-      {% endif %}
-    {% endfor %}
-
-    {% if header %}
-      <thead>
-      <tr>
-        {% for cell in header %}
-        {%
-          set cell_classes = [
-          cell.active_table_sort ? 'is-active',
-        ]
-        %}
-        <{{ cell.tag }}{{ cell.attributes.addClass(cell_classes) }}>
-        {{- cell.content -}}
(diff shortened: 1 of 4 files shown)


abel-santos-corral (dismissed):
Review finished. Only added +1 row at table.ui_patterns.yml and changed the content of the three rows of the preview.

templates/overrides/views/views-view-table.html.twig
```diff
@@ -4,97 +4,21 @@
...
+      }]) %}
     {% endfor %}
-    </tbody>
-  </table>
-</div>
+  {% endfor %}
+  {% set rows_items = rows_items|merge([{'cells': row_content}]) %}
+{% endfor %}
+{{ pattern('table', {
+  caption: caption,
```
escuriola, line 36: Missing table_responsive:
```suggestion
  caption: caption,
  table_responsive:  responsive ? true...
```

templates/overrides/views/views-view-table.html.twig
```diff
@@ -4,97 +4,21 @@
...
+  rows: rows_items,
+  table_head: table_head,
+}) }}
```
escuriola, line 40: Details (summary and description) also missing inside caption.

templates/overrides/dataset/table.html.twig
```diff
@@ -5,79 +5,9 @@
...
-          </{{ cell.tag }}>
-        {% endfor %}
-        </tr>
-      {% endfor %}
-      </tfoot>
-    {% endif %}
-  </table>
-</div>
+{{ pattern('table', {
+  caption: caption,
```
escuriola, line 9:
```suggestion
  caption: caption,
   table_responsive: true,
```

escuriola (changes requested):

templates/overrides/views/views-view-table.html.twig
escuriola: Checking with "seven" theme, seems to be a collapsible field. The summary is the link and the description the content. 
image

templates/overrides/views/views-view-table.html.twig
Maxfire: As we checked it's already in place.

templates/overrides/dataset/table.html.twig
Maxfire: same as before, it's already in place.

templates/overrides/views/views-view-table.html.twig
Maxfire: Added in the view template.

drishu (approved):

escuriola (approved):

PR #89, opened 2021-10-01
FRONT-2488: Select template with slimSelect

escuriola:
Jira issue(s):
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-424


abel-santos-corral (dismissed):
Review finished. Added a task at the bcl-builder.config.js to copy the file form_select_multiple.js.

PR #91, opened 2021-10-01
OEL-522: Drop Components module. Use ECL Twig loader

claudiu-cristea:


.drone.yml
```diff
@@ -74,7 +74,7 @@ pipeline:
     group: test
     image: registry.fpfis.eu/fpfis/httpd-php:${PHP_VERSION}-ci
     commands:
-      - ./vendor/bin/phpunit
+      - ./vendor/bin/phpunit --testdox
```
brummbar, line 77: Is this on purpose?

composer.json
```diff
@@ -7,20 +7,21 @@
...
-        "drupal/core-composer-scaffold": "^8.9 || ^9.1",
         "drupal/config_devel": "~1.2",
+        "drupal/core-composer-scaffold": "^8.9 || ^9",
         "drupal/drupal-extension": "^4.1",
         "drupal/styleguide": "^1.0@alpha",
         "drush/drush": "^10.3",
         "openeuropa/code-review": "~1.6",
-        "openeuropa/drupal-core-require-dev": "^8.9 || ^9.1",
+        "openeuropa/drupal-core-require-dev": "^8.9 || ^9",
+        "openeuropa/task-runner": "dev-master as 1.0.0-beta7",
```
brummbar, line 24: Is this needed?

.gitignore
```diff
@@ -15,7 +15,6 @@
 /.stylelintrc.json
 /docker-compose.*.yml
 /.sass-cache
-/assets/
 /.env
 /.gitattributes
 /templates/bcl
```
brummbar: This we can drop.

modules/oe_bootstrap_theme_helper/tests/Kernel/BclComponentsOverrideTest.php
```diff
@@ -0,0 +1,98 @@
+<?php
```
brummbar: This file should be in tests/src . Let's take the chance to move also the other one (MarkupRenderingTest).

modules/oe_bootstrap_theme_helper/tests/Kernel/BclComponentsOverrideTest.php
```diff
@@ -0,0 +1,98 @@
...
+      'build',
+    ];
+    new Settings($settings);
+
+    // Replace some original BCL components just to make it easy to test.
+    $path = __DIR__ . '/../../../../assets/bcl';
+    file_put_contents("$path/bcl-alert/bcl-alert.html.twig", 'Alert: oe_bootstrap_theme version');
+    file_put_contents("$path/bcl-badge/bcl-badge.html.twig", 'Badge: oe_bootstrap_theme version');
+    file_put_contents("$path/bcl-blockquote/bcl-blockquote.html.twig", 'Blockquote: oe_bootstrap_theme version');
+    file_put_contents("$path/bcl-card/bcl-card.html.twig", 'Card: oe_bootstrap_theme version');
```
brummbar: This destroys the original content of the template unfortunately. We need to resort to some other solution.

modules/oe_bootstrap_theme_helper/tests/Kernel/BclComponentsOverrideTest.php
```diff
@@ -0,0 +1,98 @@
...
+   * @dataProvider bclComponentsOverridingTestCasesProvider
+   */
+  public function testBclComponentsOverriding(string $theme, array $patterns): void {
+    // Install the theme. Subsequently, this installs also the base themes.
+    $this->container->get('theme_installer')->install([$theme]);
+    $this->config('system.theme')->set('default', $theme)->save();
+
+    foreach ($patterns as $pattern => $expected_markup) {
+      $render_array = [
+        '#type' => 'pattern',
```
brummbar: Maybe you can use `'#type' => 'pattern_preview'` to solve the issue of the comment above. So you don't have to pass any pattern data. Only the yaml will get more complex.

brummbar (changes requested):

.drone.yml
claudiu-cristea: I want to see which test runs at some point not only dots. In this way I can spot a failure and locate the failed test before the whole pipeline is finish.

composer.json
claudiu-cristea: Yes, it's for the hack in `runner.yml.dist`. Specifically, with `beta6` I'm getting:

```
Task 'exec' not supported.
```

Yeah, we need a new beta release with Task Runner, a lot of new code has accumulated in `master`.

modules/oe_bootstrap_theme_helper/tests/Kernel/BclComponentsOverrideTest.php
claudiu-cristea: We're not asserting here the content. The content of template is not relevant by any means. But how the templates are overriding each other it is. It's the main and the only goal of this test. Having simple content that identifies form which theme the template is coming is more important. Really, I would not change that for the sake of test assertion readability.

claudiu-cristea: OK, discussed in the chat. @brummbar is correct that this could break other tests, as they run on the same codebase. Will find a solution to keep the codebase unchanged.

modules/oe_bootstrap_theme_helper/tests/Kernel/BclComponentsOverrideTest.php
claudiu-cristea: I've fixed this by creating BCL components & patterns only for the purpose of testing.

PR #109, opened 2021-11-09
OEL-739: Text with feature media video variants

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-739


includes/paragraphs.inc
```diff
@@ -72,30 +75,19 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_text_feature_media(array &$
   /** @var \Drupal\paragraphs\Entity\Paragraph $paragraph */
   $paragraph = $variables['paragraph'];
 
-  // Link field doesn't exist in oe_paragraphs <= 1.8.0, so ensure it exists.
-  if ($paragraph->hasField('field_oe_link') && !$paragraph->get('field_oe_link')->isEmpty()) {
```
drishu, line 76: link is not needed ?

includes/paragraphs.inc
```diff
@@ -114,17 +106,17 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_text_feature_media(array &$
...
-  if ($source instanceof MediaAvPortalPhotoSource || $source instanceof Image) {
-    $thumbnail = $media->get('thumbnail')->first();
-    $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_bootstrap_theme_medium_no_crop');
-    $cacheability->applyTo($variables);
-    // Overwrite the variant because is an image.
-    $variables['variant'] = 'image';
+  $is_image = $source instanceof MediaAvPortalPhotoSource || $source instanceof Image;
+  $is_video = $source instanceof MediaAvPortalVideoSource || $source instanceof OEmbed || $source instanceof Iframe;
+
+  // If it's no image or no video, bail out.
```
drishu:
```suggestion
  // If it's not an image or not a video, bail out.
```

includes/paragraphs.inc
```diff
@@ -377,5 +369,79 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+      return str_replace(['_featured', '_simple'], '', $variant);
+
+    default:
+      return $variant;
+  }
+
+}
+
+/**
+ * Set the image data if media is an image.
```
drishu: add params

includes/paragraphs.inc
```diff
@@ -377,5 +369,79 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_bootstrap_theme_medium_no_crop');
+  // Overwrite the variant because is an image.
+  $variables['variant'] = 'image';
+  $cacheability->applyTo($variables);
+
+  return $variables;
+}
+
+/**
+ * Set the embedded media.
```
drishu: params

includes/paragraphs.inc
```diff
@@ -377,5 +369,79 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+  set_ratio($media, $variables, $cacheability);
+  return $variables;
+}
+
+/**
+ * Set the video iframe ratio.
+ */
+function set_ratio(Media $media, array &$variables, CacheableMetadata $cacheability): void {
+  // When dealing with iframe videos, also respect its given aspect ratio.
+  if ($media->bundle() === 'video_iframe') {
```
drishu: for more code clarity lets do it like this:
```suggestion
  if ($media->bundle() !== 'video_iframe') {
    return $variables;
  }
```

templates/patterns/featured_media/pattern-featured-media--variant-inside_text.html.twig
```diff
@@ -0,0 +1,34 @@
...
+ * @file
+ * Featured media component.
+ */
+#}
+
+{% if alignment == 'right' %}
+  {% set text_alignment_class = 'order-md-1' %}
+  {% set video_alignment_class = 'order-md-2' %}
+{% else %}
+  {% set text_alignment_class = 'order-md-2' %}
```
drishu: set these as default and after override if `alignment == 'right'` so we skip the else

drishu (changes requested):

includes/paragraphs.inc
escuriola: We have a problem with the cyclomatic complexity threshold and I had to remove as many "if" as possible. Link already comes in content so for this reason I don't need to add to variables

includes/paragraphs.inc
abel-santos-corral: In fact, must be: // If it's not an image **and** not a video, bail out.

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+
+/**
+ * Set the image data if media is an image.
+ *
+ * @param bool $is_image
+ *   Flag to determine if is an image or not.
+ * @param Drupal\media\Entity\Media $media
+ *   Media object.
+ * @param array $variables
+ *   Drupal variables.
```
abel-santos-corral: In other places is 'The render array'. Change it so is consistent everywhere, please.

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+      return str_replace(['_featured', '_simple'], '', $variant);
+
+    default:
+      return $variant;
+  }
+
+}
+
+/**
+ * Set the image data if media is an image.
```
abel-santos-corral:
```suggestion
 * Sets the image data if the media is an image.
```

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+  return $variables;
+}
+
+/**
+ * Set the embedded media.
+ *
+ * @param bool $is_image
+ *   Flag to determine if is an image or not.
+ * @param array $variables
+ *   Drupal variables.
```
abel-santos-corral: Same as per the previous comment.

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+ *   Media object.
+ * @param array $variables
+ *   Drupal variables.
+ * @param Drupal\Core\Cache\CacheableMetadata $cacheability
+ *   CacheableMetadata object.
+ *
+ * @return array
+ *   Updated variables.
+ */
+function set_ratio(Media $media, array &$variables, CacheableMetadata $cacheability): array {
```
abel-santos-corral: Please, rename it to something like set_video_iframe_ratio to make it  more explicit.

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+ *
+ * @throws \Drupal\Core\TypedData\Exception\MissingDataException
+ */
+function set_image(bool $is_image, Media $media, array &$variables, CacheableMetadata $cacheability): array {
+  if (!$is_image) {
+    return $variables;
+  }
+
+  $thumbnail = $media->get('thumbnail')->first();
+  $variables['image'] = ImageValueObject::fromStyledImageItem($thumbnail, 'oe_bootstrap_theme_medium_no_crop');
```
abel-santos-corral, line 406: Couldn't it be like date in the List Item? This would mean to get the Image style not fixed as here but to get the 'configurable' image style set at the Thumbnail field in the Manage display in http://localhost:8080/build/admin/structure/media/manage/remote_video/display

includes/paragraphs.inc
```diff
@@ -377,5 +373,124 @@ function get_bcl_variant(Paragraph $paragraph): string {
...
+ */
+function set_ratio(Media $media, array &$variables, CacheableMetadata $cacheability): array {
+  // When dealing with iframe videos, also respect its given aspect ratio.
+  if ($media->bundle() !== 'video_iframe') {
+    return $variables;
+  }
+  $ratio = $media->get('oe_media_iframe_ratio')->value;
+  $variables['ratio'] = str_replace('_', 'x', $ratio);
+  $cacheability->applyTo($variables);
+  return $variables;
```
abel-santos-corral: If setting values per reference, is necessary to return this?

templates/paragraphs/paragraph--oe-text-feature-media.html.twig
```diff
@@ -14,15 +14,19 @@
     '#alt':        image.alt,
   } %}
 {% endif %}
+{% if content.field_oe_text_long['#items'] %}
```
abel-santos-corral, line 17: The default value is set at the paragraph.inc as FALSE, can't it be all set in same place (twig or paragraph.inc)?

tests/src/Kernel/Paragraphs/MediaParagraphsTest.php
```diff
@@ -36,6 +37,237 @@ protected function setUp(): void {
     media_install();
   }
 
+  /**
+   * Test 'text with featured media' paragraph rendering.
+   */
+  public function testFeaturedMedia(): void {
+    $image_file = file_save_data(file_get_contents(drupal_get_path('theme', 'oe_bootstrap_theme') . '/tests/fixtures/example_1.jpeg'), 'public://example_1_en.jpeg');
```
abel-santos-corral: If the file is not needed as _en, use same name as source file.

abel-santos-corral (changes requested):

includes/paragraphs.inc
escuriola: Is needed because of the sign of the method: 

`function set_video_iframe_ratio(Media $media, array &$variables, CacheableMetadata $cacheability): array {`

is waiting an array as the output of the method. So the return is needed for this reason.

templates/paragraphs/paragraph--oe-text-feature-media.html.twig
escuriola: No because this settings deppends of the full text field is fulfilled or not, so by default is false but if the full text field has some content, the settings must be true.

includes/paragraphs.inc
escuriola: I'm not sure of this. I'm using here the oe_theme way, maybe another way is possible, we can check this together.

drishu (approved):

Maxfire (approved):

abel-santos-corral (approved):

PR #111, opened 2021-11-16
OEL-764: Fix white text when image is shade.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-764


abel-santos-corral:
this was meant for version 11. Now we're developing for 13, this becomes obsolete. Closing PR.

PR #113, opened 2021-11-17
OEL-750: Upgrade BCL to 0.14.0

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-750

Includes Branches|Issues:
- Branch OEL-750: Includes BCL Patterns 0.14.0
- Card [OEL-751](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-751)
- Banner: [OEL-752](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-752)
- Listing pattern: [OEL-757](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-757)
- Date block composition: [OEL-760](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-760)
- Link: [OEL-761](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-761)
- Search Card variant: [OEL-763](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-763)

### DIFF
--- a/templates/patterns/banner/banner.ui_patterns.yml
+++ b/templates/patterns/banner/banner.ui_patterns.yml
@@ -38,10 +38,12 @@ banner:
       description: "Heading of the banner"
       preview: "EU Budget for the future"
     description:
-      type: "text"
+      type: "array"
       label: "Description"
-      description: "Sub-heading of the banner"
-      preview: "Innovation, economy, environment and geopolitics"
+      description: "Sub-heading of the banner [content, classes]"
+      preview:
+        - content: "Innovation, economy, environment and geopolitics"
+          classes: "my-2"
     link:
       type: "array"
       label: "Link"
@@ -60,8 +62,3 @@ banner:
       label: "Image"
       description: "URL to the image"
       preview:  "https://placeimg.com/1000/500/arch"
-    content_classes:
-      type: "text"
-      label: "Content classes"
-      description: "CSS classes to add to the content container"
-      preview: "my-2"
--- a/templates/patterns/banner/banner.ui_patterns.yml
+++ b/templates/patterns/banner/banner.ui_patterns.yml
@@ -58,3 +58,7 @@ banner:
       type: "text"
       label: "Image"
       preview:  "https://placeimg.com/1000/500/arch"
+    content_classes:
+      type: "text"
+      label: "Content classes"
+      preview: "my-2"
--- a/templates/patterns/banner/banner.ui_patterns.yml
+++ b/templates/patterns/banner/banner.ui_patterns.yml
@@ -38,12 +38,10 @@ banner:
       description: "Heading of the banner"
       preview: "EU Budget for the future"
     description:
-      type: "array"
+      type: "text"
       label: "Description"
-      description: "Sub-heading of the banner [content, classes]"
-      preview:
-        - content: "Innovation, economy, environment and geopolitics"
-          classes: "my-2"
+      description: "Sub-heading of the banner"
+      preview: "Innovation, economy, environment and geopolitics"
     link:
       type: "array"
       label: "Link"
--- a/templates/patterns/banner/banner.ui_patterns.yml
+++ b/templates/patterns/banner/banner.ui_patterns.yml
@@ -1,6 +1,6 @@
 banner:
-  label: "Banner composition"
-  description: "Hidden sidebar with activation. https://getbootstrap.com/docs/5.0/components/offcanvas/"
+  label: "Banner"
+  description: "Banner or heroes with text and/or background image and/or with call to action. https://getbootstrap.com/docs/5.0/examples/heroes/"
   variants:
     default:
       label: Default text banner
@@ -14,6 +14,7 @@ banner:
     shade:
       type: "boolean"
       label: "Shade"
+      description: "Enable the shadow effect to the image."
       preview: false
     hero:
       type: "boolean"
@@ -57,8 +58,10 @@ banner:
     image:
       type: "text"
       label: "Image"
+      description: "URL to the image"
       preview:  "https://placeimg.com/1000/500/arch"
     content_classes:
       type: "text"
       label: "Content classes"
+      description: "CSS classes to add to the content container"
       preview: "my-2"
--- a/templates/patterns/banner/pattern-banner.html.twig
+++ b/templates/patterns/banner/pattern-banner.html.twig
@@ -15,16 +15,14 @@
   {% set attributes = attributes.addClass(['text-center'])%}
 {% endif %}
 {% include '@oe-bcl/bcl-banner/banner.html.twig' with {
-  shade: shade,
+  shade: variant == 'image_shade' ? true : shade,
   hero: hero,
   full_width: full_width,
   centered: centered,
   title: title,
-  description: description,
+  description: description.content,
   link: link,
-  image: image,
-  content_classes: content_classes,
-} %}
-
-
-
+  image: variant == 'image' or variant == 'image_shade' ? image : '' ,
+  content_classes: description.classes,
+  attributes: attributes,
+} only %}
--- a/templates/patterns/banner/pattern-banner.html.twig
+++ b/templates/patterns/banner/pattern-banner.html.twig
@@ -3,61 +3,28 @@
  * @file
  * Banner pattern.
  */
-Parameters:
-  - title: (string) (default: "")
-  - description: (string) (default: "")
-  - link: (link object) (default: {}): predefined structure for Link component
-  - image: (string) <img> tag
-  - shade: (boolean) (default: false)
-  - hero: (boolean) (default: false)
-  - full_width: (boolean) (default: false)
 #}
 {% block attributes %}
-{% set attributes = attributes.addClass([
-    'bg-light',
-    'text-dark',
-  ])
-%}
+  {% set attributes = attributes.addClass([
+      'bg-lighter',
+      'text-dark',
+    ])
+  %}
 {% endblock %}
 {% if centered  %}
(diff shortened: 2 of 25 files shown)


templates/patterns/banner/banner.ui_patterns.yml
```diff
@@ -57,4 +58,10 @@ banner:
     image:
       type: "text"
       label: "Image"
+      description: "URL to the image"
       preview:  "https://placeimg.com/1000/500/arch"
+    content_classes:
+      type: "text"
+      label: "Content classes"
+      description: "CSS classes to add to the content container"
+      preview: "my-2"
```
brummbar: We shouldn't expose presentation details like classes in the pattern, we need to offer a layer to pass data only.

templates/patterns/date_block/date_block.ui_patterns.yml
```diff
@@ -0,0 +1,24 @@
+date_block:
+  label: "Date Block"
```
brummbar: Don't capitalise words.

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -19,6 +19,12 @@ link:
     disabled:
       type: boolean
       label: Disabled
+      description: Disable the link.
+    standalone:
+      type: boolean
+      label: Disabled
```
brummbar: Wrong label.

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -19,6 +19,12 @@ link:
...
+      description: Add extra "text-underline-hover" class to link item.
```
brummbar: Explain what it does, not the technical detail (class name).

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -28,9 +34,11 @@ link:
     remove_icon_spacers:
       type: boolean
       label: Remove icon spacers
+      description: When enabled, not add the margin class to the icon
```
brummbar:
```suggestion
      description: When set to true, prevents margins to be added to the icon.
```

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -28,9 +34,11 @@ link:
...
     link_id:
       type: textfield
       label: Button id
+      description: Add id attribute to the link.
```
brummbar:
```suggestion
      description: A specific ID to use for the link.
```
Not sure this should be left at all.

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
```diff
@@ -17,9 +17,42 @@ default_page_banner_centered_with_link_full_widht:
...
-      'div.bg-light.text-dark.bcl-banner': 1
+      'div.bg-lighter.text-dark.bcl-banner': 1
       'div.full-width': 1
       'div.text-center': 1
+      'div.bcl-banner__content': 1
+      'a': 1
+      'div.hero': 0
+      'div.bcl-banner__image': 0
+    contains:
+      'bcl-banner__content': "EU Budget for the future"
```
brummbar: This is not a valid selector, but tests are green, which makes me wonder.

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -28,9 +34,11 @@ link:
     remove_icon_spacers:
       type: boolean
       label: Remove icon spacers
+      description: When enabled, not add the margin class to the icon
     link_id:
       type: textfield
       label: Button id
```
brummbar: This is not a button though.

templates/patterns/banner/pattern-banner.html.twig
```diff
@@ -3,61 +3,28 @@
...
-  {% set _full_width = full_width|default(false) %}
+{% include '@oe-bcl/bcl-banner/banner.html.twig' with {
+  shade: shade,
+  hero: hero,
+  full_width: full_width,
+  centered: centered,
+  title: title,
+  description: description,
+  link: link,
+  image: image,
```
brummbar: By looking at the components page, I see that the image is shown in all the variants. We shouldn't pass the image if we are rendering a variant where is not needed. It seems it's not handled on the BCL side.

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -71,10 +71,11 @@ card:
       label: "Image"
       description: "Card image [path, alt, position]"
       preview:
-        path: "https://placeimg.com/1000/500/arch"
+        path: "https://placeimg.com/1000/500/people"
         alt: "Alternative text for card image"
         position: "top"
         rounded: 3
+        '#markup': "https://placeimg.com/1000/500/tech"
```
brummbar: Why this line?

templates/patterns/listing/listing.ui_patterns.yml
```diff
@@ -21,7 +38,7 @@ listing:
       preview:
         - title:
             content:
-              '#markup' : '<a href="/example">Card title 1</a>'
+              '#markup' : '<a class="text-underline-hover" href="/example">Card title 1</a>'
```
brummbar: Why this change? I don't think the class is needed.

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -114,3 +115,6 @@ card:
         - label: "Secondary badge"
           assistive_text: "Assistive text for secondary badge"
           rounded_pill: true
+    date:
+      type: "array"
+      label: "Date that will be shown with Date Block."
```
brummbar:
```suggestion
      label: 'Date that will be rendered with the "Date block" pattern.'
```

templates/patterns/date_block/date_block.ui_patterns.yml
```diff
@@ -0,0 +1,24 @@
...
+    year:
+      type: "number"
+      label: "Year"
+      description: "The year number in format AAAA."
+      preview: "2021"
+    date_time:
+      type: "text"
+      label: "Datetime"
+      description: "The datetime attribute for the block which allows translation of dates into machine-readable format."
+      preview: "2021-05-27"
```
brummbar: I don't like this much. We are forced to specify the data twice to be passed in the pattern. Instead, we can simply accept one single parameter, like a DrupalDateTime or a formatted date, and then in a preprocess elaborate the various pieces of the date.

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
brummbar: I check and there are around 50 failures due to the selector being ignored, which is not good. I'll create another ticket for this.

templates/patterns/banner/pattern-banner.html.twig
```diff
@@ -3,61 +3,28 @@
...
+  content_classes: content_classes,
+} %}
```
brummbar: We need to pass only the expected parameters.
```suggestion
} only %}
```

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
```diff
@@ -17,9 +17,42 @@ default_page_banner_centered_with_link_full_widht:
...
+      'p': "Innovation, economy, environment and geopolitics"
```
brummbar: Let's already put here some proper selectors though.

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
```diff
@@ -17,9 +17,42 @@ default_page_banner_centered_with_link_full_widht:
...
+    equals:
+      'a': "Subscribe"
```
brummbar: This is way too generic, we need to assert that we are rendering using the correct bcl template. At least something like `.bcl-banner__content a[href="https://example.com"]: Subscribe`.

brummbar (dismissed):
I have many changes to introduce but I cannot add them all here. Let's try to move forward with this and we rework patterns in a separate ticket.

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
escuriola: @brummbar contains assertion selectors are not used, only checks the texts exists on the html:
```
// Assert presence of given strings.
    foreach ($assertions['contains'] as $string) {

      $this->assertStringContainsString($string, $html);

    }
```

Sometimes we use not selector and use for instance anchor_1, anchor_2, and so on...

templates/patterns/card/card.ui_patterns.yml
escuriola: Because of the search variant of the card. Has a special way to get the image.

templates/patterns/banner/banner.ui_patterns.yml
escuriola: @brummbar @drishu We would need an example of how to manage this change. Convert "description" to "array" or to "render" and then prepare the internal content_classes field from there? Is there any example where we can review this?

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
escuriola: Are you agree to manage all this in another ticket @brummbar @drishu ? Can I create the Jira ticket to the next sprint?

drishu (approved):
Some improvements will be needed but we agreed to follow up

timpevi (approved):

Maxfire (approved):

PR #123, opened 2021-11-22
OEL-759: Theme the Facts & Figures paragraph type

abel-santos-corral:
Jira issue(s):
- [OEL-759](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-759)
- [OEL-847](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-847) --> resolves _core/config_hash problem


modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_variant.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
```
escuriola: This more than a variant is a setting and the variant is something internal. Let's rename this field.

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_disp_icons.yml
```diff
@@ -0,0 +1,22 @@
...
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_disp_icons
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_disp_icons
+field_name: oe_bt_facts_figures_disp_icons
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Display icons'
+description: ''
```
escuriola: Add description for the new fields

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_equal_height.yml
```diff
@@ -0,0 +1,22 @@
...
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_equal_height
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_equal_height
+field_name: oe_bt_facts_figures_equal_height
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Equal height'
+description: ''
```
escuriola: Same above

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns.yml
```diff
@@ -0,0 +1,24 @@
...
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_grid_columns
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns
+field_name: oe_bt_facts_figures_grid_columns
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Grid columns'
+description: ''
```
escuriola: Same above

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_resp_cols.yml
```diff
@@ -0,0 +1,24 @@
...
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_resp_cols
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_resp_cols
+field_name: oe_bt_facts_figures_resp_cols
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Responsive columns'
+description: ''
```
escuriola: Same above

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.info.yml
```diff
@@ -0,0 +1,23 @@
...
+core_version_requirement: ^8.9 || ^9.1
+dependencies:
+  - oe_paragraphs:oe_paragraphs
+  - oe_bootstrap_theme:oe_bootstrap_theme_helper
+config_devel:
+  install:
+    - field.field.paragraph.oe_links_block.oe_bt_links_block_background
+    - field.field.paragraph.oe_links_block.oe_bt_links_block_direction
+    - field.field.paragraph.oe_social_media_follow.oe_bt_links_block_background
+    - field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_disp_icons
```
escuriola: Missing equal_heights here

templates/patterns/fact_figures/fact_figures.ui_patterns.yml
```diff
@@ -0,0 +1,103 @@
...
+          subtitle: "Jira ticket"
+          title: "345 Dev Ticket"
+          description: "Class aptent taciti sociosqu ad litora torquent per conubia nostra, per inceptos himenaeos. Duis nec lectus tortor. "
+        - icon:
+            name: "eye-fill"
+            size: "l"
+          subtitle: "Figma components"
+          title: "43 Components"
+          description: "Sed efficitur bibendum rutrum. Nunc feugiat congue augue ac consectetur."
+    cols_extra_classes:
```
escuriola: Remove classes from pattern definition.

templates/paragraphs/paragraph--oe-facts-figures.html.twig
```diff
@@ -0,0 +1,22 @@
...
+{{ pattern('fact_figures', {
+  'style': style,
+  'title': title,
+  'items': items,
+  'display_icons': display_icons,
+  'responsive_columns': responsive_columns,
+  'grid_columns': grid_columns,
+  'gutter': gutter,
+  'equal_height': equal_height,
+  'cols_extra_classes': cols_extra_classes,
```
escuriola: cols_extra_classes will be no longer a field of the pattern

templates/patterns/fact_figures/pattern-fact-figures.html.twig
```diff
@@ -0,0 +1,20 @@
...
+  'display_icons': display_icons,
+  'responsive_columns': responsive_columns,
+  'grid_columns': grid_columns,
+  'gutter': gutter,
+  'equal_height': equal_height,
+  'cols_extra_classes': cols_extra_classes,
+  'responsiveness': responsiveness,
+  'icons_path': icons_path,
+  'link_more': link_more,
+} %}
```
escuriola:
```suggestion
} only %}
```

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_equal_height.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
```
escuriola: Something extrange with the machine name of this field in my installation: 
image

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_view_display.paragraph.oe_facts_figures.default.yml
```diff
@@ -0,0 +1,97 @@
...
+    type: number_integer
+    region: content
+  oe_bt_facts_figures_variant:
+    weight: 3
+    label: above
+    settings: {  }
+    third_party_settings: {  }
+    type: list_default
+    region: content
+hidden: {  }
```
escuriola: Settings fields must be hidden on the view display

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+  }
+  $ratio = $media->get('oe_media_iframe_ratio')->value;
+  $variables['ratio'] = str_replace('_', 'x', $ratio);
+  $cacheability->applyTo($variables);
+}
+
+/**
+ * Implements hook_preprocess_paragraph().
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_facts_figures(array &$variables): void {
```
escuriola: Move this with all  preprocess paragraphs function to be all together, let the auxiliary function to the bottom of the file

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+  $paragraph = $variables['paragraph'];
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value;
```
escuriola:
```suggestion
    $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
```

escuriola: Remove the if

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+  foreach ($paragraph->get('field_oe_paragraphs')->referencedEntities() as $sub_paragraph) {
+    // Get sub-paragraph translation.
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    $variables['items'][] = [
+      'icon' => [
+        'name' => $sub_paragraph->get('field_oe_icon')->value,
+        'size' => "l",
+      ],
+      'title' => $sub_paragraph->get('field_oe_title')->value,
```
escuriola:
```suggestion
      'title' => $sub_paragraph->get('field_oe_title')->value ?? '',
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+      'subtitle' => $sub_paragraph->get('field_oe_subtitle')->value,
```
escuriola:
```suggestion
      'subtitle' => $sub_paragraph->get('field_oe_subtitle')->value ?? '',
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+      'description' => $sub_paragraph->get('field_oe_plain_text_long')->value,
```
escuriola:
```suggestion
      'description' => $sub_paragraph->get('field_oe_plain_text_long')->value ?? '',
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+      'rss' => 'RSS',
+      'search' => 'Search',
+      'share' => 'Share',
+      'file-spreadsheet' => 'Spreadsheet',
+      'twitter' => 'Twitter',
+      'person-video' => 'Video',
+    ];
+    // Change the label of the empty option.
+    $options = [];
+    foreach ($options_field_icons as $key => $value) {
```
escuriola: We don't need this foreach

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -0,0 +1,591 @@
...
+
+}
+
+/**
+ * Implements hook_options_list_alter().
+ */
+function oe_bootstrap_theme_paragraphs_options_list_alter(array &$options, array $context): void {
+  // Check if this is the field we want to change.
+  if ($context['fieldDefinition']->id() == 'paragraph.oe_fact.field_oe_icon') {
+    $options_field_icons = [
```
escuriola: rename to options

tests/src/Functional/ParagraphsTest.php
```diff
@@ -0,0 +1,213 @@
...
+   * {@inheritdoc}
+   */
+  protected $defaultTheme = 'stark';
+
+  /**
+   * The administration theme name.
+   *
+   * @var string
+   */
+  protected $adminTheme = 'stark';
```
escuriola: We need this?

tests/src/Functional/ParagraphsTest.php
```diff
@@ -0,0 +1,213 @@
...
+    $this->assertSession()->pageTextContains('Read more');
+    $this->assertSession()->pageTextContains('1529 JIRA Ticket');
+    $this->assertSession()->pageTextContains('Jira Tickets');
+    $this->assertSession()->pageTextContains('Nunc condimentum sapien ut nibh finibus suscipit vitae at justo. Morbi quis odio faucibus, commodo tortor id, elementum libero.');
+  }
+
+  /**
+   * Create content type with paragraphs field.
+   */
+  protected function createContentTypeTestWithParagraphs(): void {
```
escuriola: rename to addParagraphFieldToContentType

escuriola (dismissed):
Missing also a FactsAndFiguresTest.php in test > src > Kernel > Paragraphs

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_variant.yml
abel-santos-corral: Rename it as Layout

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_equal_height.yml
abel-santos-corral: @escuriola I reviewed at my local and is working fine. I reinstalled a fresh site, and is there with complete name. Let's have a call once I deliver the modifications for your review and check together if this still happens. I had some issues with PHPStorm(TM) the other day, I fixed just to not have that horrible name for the field. So it has to be solved.

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_disp_icons.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
```
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_equal_height.yml
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns.yml
```diff
@@ -0,0 +1,24 @@
...
+description: 'Sets the number of grid columns. Minimum number is 1 column.'
+required: false
+translatable: false
+default_value:
+  -
+    value: 1
+default_value_callback: ''
+settings:
+  min: 1
+  max: null
```
drishu: max 3

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_resp_cols.yml
```diff
@@ -0,0 +1,24 @@
+langcode: en
```
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns.yml
```diff
@@ -0,0 +1,24 @@
...
+status: true
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_grid_columns
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns
+field_name: oe_bt_facts_figures_grid_columns
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Grid columns'
```
drishu: In Listing item block we call this grid system "Layout", lets keep consistency, also the values are same, 1, 2 and 3, so maybe we can reuse that field storage

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_layout.yml
```diff
@@ -0,0 +1,22 @@
...
+  config:
+    - field.storage.paragraph.oe_bt_facts_figures_layout
+    - paragraphs.paragraphs_type.oe_facts_figures
+  module:
+    - options
+id: paragraph.oe_facts_figures.oe_bt_facts_figures_layout
+field_name: oe_bt_facts_figures_layout
+entity_type: paragraph
+bundle: oe_facts_figures
+label: Layout
```
drishu: In Links Block we called this "Background", lets do the same here

modules/oe_bootstrap_theme_paragraphs/config/install/field.storage.paragraph.oe_bt_facts_figures_disp_icons.yml
```diff
@@ -0,0 +1,17 @@
+langcode: en
```
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/install/field.storage.paragraph.oe_bt_facts_figures_equal_height.yml
```diff
@@ -0,0 +1,17 @@
+langcode: en
```
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/install/field.storage.paragraph.oe_bt_facts_figures_resp_cols.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
```
drishu: remove this

modules/oe_bootstrap_theme_paragraphs/config/overrides/field.storage.paragraph.field_oe_icon.yml
```diff
@@ -0,0 +1,86 @@
+langcode: en
```
drishu: rather then this, use the old field, but use allowed_values, see https://github.com/digit-devs/echo-ucpkn-dev/pull/89/files#diff-07f0b059e49e7a9dd874a2a11048b31d57a68c07da20ff2b2c2c140dfa83121dR16

The idea is to do this on oe_paragraph side, the callback should send out an event and implement a subscriber, the subscriber will return the values present now in oe_paragraphs.
Then in this PR we implement our subscriber and override the values with ours.
Since we change oe_parapgraphs a ewmcs PR should be provided to check the result.
Create a new ticket for this if needed.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.install
```diff
@@ -29,6 +29,9 @@ function oe_bootstrap_theme_paragraphs_install($is_syncing): void {
     'core.entity_view_display.paragraph.oe_links_block.default',
     'core.entity_view_display.paragraph.oe_social_media_follow.default',
     'field.field.paragraph.oe_social_media_follow.field_oe_social_media_variant',
+    'core.entity_form_display.paragraph.oe_facts_figures.default',
+    'core.entity_view_display.paragraph.oe_facts_figures.default',
+    'field.storage.paragraph.field_oe_icon',
```
drishu: If I have the module installed then I don't get the fields, pls check if ucpkn has the module enabled and provide update if yes

drishu (changes requested):

modules/oe_bootstrap_theme_paragraphs/config/install/field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_grid_columns.yml
abel-santos-corral: As discussed, the field of Layout of List item block is present at oe_paragraphs. We can't use it. So the best solution will be to rename the field at Facts and figure and leave it dispo for forthcoming paragraphs.

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsAlterSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+ * Provides options for the icon field.
+ */
+class IconOptionsAlterSubscriber implements EventSubscriberInterface {
+
+  /**
+   * {@inheritdoc}
+   */
+  public static function getSubscribedEvents(): array {
+    return [
+      IconOptionsEvent::class => ['getIconOptions', 300],
```
drishu:
```suggestion
      IconOptionsEvent::class => ['getIconOptions', -1],
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.install
```diff
@@ -19,30 +19,37 @@ function oe_bootstrap_theme_paragraphs_install($is_syncing): void {
   if ($is_syncing) {
     return;
   }
-  $storage = new FileStorage(drupal_get_path('module', 'oe_bootstrap_theme_paragraphs') . '/config/overrides/');
-  $config_manager = \Drupal::service('config.manager');
-  $entity_type_manager = \Drupal::entityTypeManager();
 
-  $field_config = [
+  $folder = '/config/overrides/';
```
drishu: $folder is used only below, so no need to define it as a variable, remove this line

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.install
```diff
@@ -19,30 +19,37 @@ function oe_bootstrap_theme_paragraphs_install($is_syncing): void {
...
-
-    $entity_type = $config_manager->getEntityTypeIdByName($name);
-    /** @var \Drupal\Core\Config\Entity\ConfigEntityStorageInterface $entity_storage */
-    $entity_storage = $entity_type_manager->getStorage($entity_type);
-    $id_key = $entity_storage->getEntityType()->getKey('id');
-    $entity = $entity_storage->load($configuration[$id_key]);
-    $entity = $entity_storage->updateFromStorageRecord($entity, $configuration);
-    $entity->save();
-  }
+  ConfigUtil::importConfigFromFile($config, $folder);
```
drishu:
```suggestion
  ConfigUtil::importConfigFromFile($config, '/config/overrides/');
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.install
```diff
@@ -19,30 +19,37 @@ function oe_bootstrap_theme_paragraphs_install($is_syncing): void {
...
+}
+
+/**
+ * Upgrade the Facts and Figures paragraph.
+ */
+function oe_bootstrap_theme_update_8001(): void {
```
drishu: lets stick to schema changes in hook_update_N and use post_update for config changes

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.install
```diff
@@ -19,30 +19,37 @@ function oe_bootstrap_theme_paragraphs_install($is_syncing): void {
...
+    'field.storage.paragraph.oe_bt_layout',
+    'field.field.paragraph.oe_facts_figures.oe_bt_facts_figures_background',
+    'field.field.paragraph.oe_facts_figures.oe_bt_layout',
+  ];
+
+  $config['displays'] = [
+    'core.entity_form_display.paragraph.oe_facts_figures.default',
+    'core.entity_view_display.paragraph.oe_facts_figures.default',
+  ];
+  ConfigUtil::importConfigFromFile($config, $folder);
```
drishu: use a static string instead of $folder

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
   $variables['hide_from'] = $paragraph->get('field_oe_timeline_expand')->value;
 }
 
+/**
+ * Implements hook_preprocess_paragraph().
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_facts_figures(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+  $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
```
drishu: we should be using isEmpty(), change this everywhere you find it

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+
+  if (!$paragraph->get('field_oe_link')->isEmpty()) {
+    $variables['link_more']['path'] = $paragraph->get('field_oe_link')->first()->getUrl()->toString();
+    $variables['link_more']['label'] = $paragraph->get('field_oe_link')->first()->get('title')->getValue();
+  }
+  $variables['items'] = [];
+
+  /** @var \Drupal\paragraphs\Entity\Paragraph $sub_paragraph */
+  foreach ($paragraph->get('field_oe_paragraphs')->referencedEntities() as $sub_paragraph) {
+    // Get sub-paragraph translation.
```
drishu: you can name a var sub-something but lets not define new terms
```suggestion
    // Get the paragraph translation.
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    $variables['items'][] = [
+      'icon' => [
+        'name' => str_replace('_', '-', $sub_paragraph->get('field_oe_icon')->value),
+        'size' => "l",
```
drishu: dont mix single and double quotes for nothing

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+      ],
+      'title' => $sub_paragraph->get('field_oe_title')->value ?? '',
```
drishu: use isEmpty in call cases

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+      'subtitle' => $sub_paragraph->get('field_oe_subtitle')->value ?? '',
+      'description' => $sub_paragraph->get('field_oe_plain_text_long')->value ?? '',
+    ];
+  }
+
+  $variables['equal_height'] = TRUE;
```
drishu: set default config in the pattern template, not in the preprocess

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +329,48 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+  $variables['display_icons'] = TRUE;
+
+  if (!$paragraph->get('oe_bt_facts_figures_background')->isEmpty()) {
```
drishu: yes, this is the proper way to do it!

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.services.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  oe_bootstrap_theme_paragraphs.icon_options_subscriber:
+    class: 'Drupal\oe_bootstrap_theme_paragraphs\EventSubscriber\IconOptionsAlterSubscriber'
```
drishu:
```suggestion
    class: 'Drupal\oe_bootstrap_theme_paragraphs\EventSubscriber\IconOptionsSubscriber'
```

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsAlterSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+
+namespace Drupal\oe_bootstrap_theme_paragraphs\EventSubscriber;
+
+use Drupal\oe_paragraphs\Event\IconOptionsEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Provides options for the icon field.
+ */
+class IconOptionsAlterSubscriber implements EventSubscriberInterface {
```
drishu: we should name it the way it makes most sense, which for me is IconOptionsSubscriber, no need to find a different name just to differentiate that's what we have namespaces for

src/ConfigUtil.php
```diff
@@ -0,0 +1,64 @@
+<?php
+
+namespace Drupal\oe_bootstrap_theme;
+
+use Drupal\Core\Config\FileStorage;
+use Drupal\Component\Utility\Crypt;
+
+/**
+ * Utility class with static methods related to configuration stuff.
```
drishu: Don't describe what it has (so don't mention the static method), describe what its for.
Also, don't use "stuff", that helps with nothing to describe what this is for.

src/ConfigUtil.php
```diff
@@ -0,0 +1,64 @@
...
+ */
+class ConfigUtil {
+
+  /**
+   * Installs configuration for paragraphs from file.
```
drishu: Except "oe_bootstrap_theme_paragraphs" there is not hint to paragraphs, we could use this for any config if we jsut remove that bit and send it as a parameter, so lets add here a more generic description.

src/ConfigUtil.php
```diff
@@ -0,0 +1,64 @@
...
+   *
+   * @param array $configuration_names
+   *   Contains the name of configuration files that must be installed.
+   * @param string $folder
+   *   Contains the folder where the config resides.
+   */
+  public static function importConfigFromFile(array $configuration_names, string $folder): void {
```
drishu:
```suggestion
  public static function importConfigFromFile(string $module, string $path, array $config_names): void {
```

src/ConfigUtil.php
```diff
@@ -0,0 +1,64 @@
...
+    $storage = new FileStorage(drupal_get_path('module', 'oe_bootstrap_theme_paragraphs') . $folder);
```
drishu:
```suggestion
    $storage = new FileStorage(drupal_get_path('module', $module) . $folder);
```

src/ConfigUtil.php
```diff
@@ -0,0 +1,64 @@
...
+    $config_manager = \Drupal::service('config.manager');
+    $entity_type_manager = \Drupal::entityTypeManager();
+
+    if (isset($configuration_names['fields'])) {
+      // Clear the field widget cache.
```
drishu: its clear what this does

templates/paragraphs/paragraph--oe-facts-figures.html.twig
```diff
@@ -0,0 +1,21 @@
...
+#}
+{{ pattern('fact_figures', {
+  'style': style,
+  'title': title,
+  'items': items,
+  'display_icons': display_icons,
+  'responsive_columns': responsive_columns,
+  'grid_columns': grid_columns,
+  'gutter': gutter,
+  'equal_height': equal_height,
```
drishu: don't pass these items where you added defaults, this needs to be handled in the pattern template

templates/patterns/fact_figures/fact_figures.ui_patterns.yml
```diff
@@ -0,0 +1,99 @@
...
+  settings:
+    style:
+      type: select
+      label: Style
+      description: "Style of the composition"
+      options:
+        default: default
+        block: block
+      preview: default
+    display_icons:
```
drishu: remove these settings where we have default values

templates/patterns/fact_figures/fact_figures.ui_patterns.yml
```diff
@@ -0,0 +1,99 @@
...
+      description: "Title of the block"
+      preview: "Fact and figures block"
+    items:
+      type: "array"
+      label: "List items"
+      description: "List of items"
+      preview:
+        - icon:
+            name: "x-diamond-fill"
+            size: "l"
```
drishu: we should send only the icon name
```suggestion
            - icon: "x-diamond-fill"
```

templates/patterns/fact_figures/pattern-fact-figures.html.twig
```diff
@@ -0,0 +1,20 @@
...
+#}
+{% include '@oe-bcl/bcl-fact-figures/fact-figures.html.twig' with {
+  'variant': style,
+  'title': title,
+  'items': items,
+  'display_icons': display_icons,
+  'responsive_columns': responsive_columns,
+  'grid_columns': grid_columns,
+  'gutter': gutter,
+  'equal_height': equal_height,
```
drishu: add defaults here
```suggestion
  'equal_height': true,
```

tests/src/Functional/ParagraphsTest.php
```diff
@@ -152,7 +152,85 @@ public function testSocialMediaFollowParagraph(): void {
...
 
   /**
-   * Adds a field to a paragraph.
+   * Test Facts and figures paragraphs form.
+   */
+  public function testFactsFiguresParagraph(): void {
+    $this->drupalGet('/node/add/paragraphs_test');
+    $page = $this->getSession()->getPage();
+    $page->pressButton('Add Facts and figures');
+    // Assert theFacts and figures fields appears.
```
drishu:
```suggestion
    // Assert the Facts and figures fields are present.
```

tests/src/Functional/ParagraphsTest.php
```diff
@@ -152,7 +152,85 @@ public function testSocialMediaFollowParagraph(): void {
...
+    ];
+    foreach ($allowed_values as $allowed_value) {
+      $this->assertSession()->elementsCount('css', 'option[value="' . $allowed_value . '"]', 1);
+    }
+    $this->assertSession()->elementsCount('css', 'select#edit-oe-bt-paragraphs-0-subform-field-oe-icon option', 20);
+
+  }
+
+  /**
+   * Create content type with paragraphs field.
```
drishu: this doesn't create a content type (this looks familiar, didn't we do the same in a different PR? please see also my comments from there)

tests/src/Kernel/fixtures/markup_rendering_patterns/fact_figures.yml
```diff
@@ -0,0 +1,348 @@
+fact_figures_default_1col:
```
drishu, line 1: update these once you changed the pattern

drishu (changes requested):

tests/src/Functional/ParagraphsTest.php
abel-santos-corral: No @drishu , not same, this is the Facts and Figures paragraphs, and just after that test there's one to check that the event subscriber works ok in this module oe_bootstrap_theme_paragraphs. I did same testing strategy: that the items are loaded correctly in the event subscriber and that the number of icons that are loaded are just the ones we want (checked in a count assertion).

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.post_update.php
```diff
@@ -0,0 +1,33 @@
...
+ */
+
+declare(strict_types = 1);
+
+use Drupal\oe_bootstrap_theme\ConfigUtil;
+
+/**
+ * Upgrade the Facts and Figures paragraph to include new fields.
+ *
+ * @param array $sandbox
```
drishu: hook_update_N and hook_post_name dont need more then a description

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.post_update.php
```diff
@@ -0,0 +1,33 @@
...
+ *   Stores information for batch updates.
+ */
+function oe_bootstrap_theme_paragraphs_post_update_10001(array &$sandbox) {
```
drishu:
```suggestion
function oe_bootstrap_theme_paragraphs_post_update_10001(array &$sandbox): void {
```

src/ConfigUtil.php
```diff
@@ -0,0 +1,65 @@
...
+        /* When we create a new config, it usually means that we are also ship-
+        ping it in the config/install folder, so we must ensure it gets the hash
+        so Drupal treats it as a shipped conf. This means that it gets exposed
+        to be translated via the locale system as well. */
+        $configuration['_core']['default_config_hash'] = Crypt::hashBase64(serialize($configuration));
+        $entity = $entity_storage->updateFromStorageRecord($entity, $configuration);
+        $entity->save();
+      }
+    }
+
```
drishu: remove this newline

templates/paragraphs/paragraph--oe-facts-figures.html.twig
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @file
+ * Theme override to display the 'Facts and Figures' #type paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{{ pattern('fact_figures', {
+  'style': style,
+  'title': title|default(),
```
drishu: pls test this default, I've never seen it used like this (not sending a value to the function)

src/ConfigUtil.php
```diff
@@ -0,0 +1,65 @@
...
+   *
+   * @param string $module
+   *   Contains the name of the module whose configuration will be installed.
+   * @param string $path
+   *   Contains the folder where the config resides.
+   * @param array $config_names
+   *   Contains the name of configuration files that must be installed.
+   */
+  public static function importConfigFromFile(string $module, string $path, array $config_names): void {
+    $storage = new FileStorage(drupal_get_path('module', $module) . $path);
```
drishu: we should use extension.list service instead of drupal_get_path

tests/src/Functional/ParagraphsTest.php
drishu: ok but the addParagraphsField method doesn't create a content type right ?

drishu (changes requested):
Don't forget to use isEmpty in the entire .module file instead of ??

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.post_update.php
```diff
@@ -36,3 +36,17 @@ function oe_bootstrap_theme_paragraphs_post_update_00002(array &$sandbox): void
...
+ */
+function oe_bootstrap_theme_paragraphs_post_update_00003(array &$sandbox): void {
+  $configs = [
+    'field.storage.paragraph.oe_bt_layout',
+    'field.field.paragraph.oe_facts_figures.oe_bt_layout',
+    'core.entity_form_display.paragraph.oe_facts_figures.default',
+    'core.entity_view_display.paragraph.oe_facts_figures.default',
+  ];
+
+  ConfigImporter::importMultiple('oe_bootstrap_theme_paragraphs', '/config/post_updates/00003/', $configs, TRUE);
```
drishu, line 51: Now that we have this paragraph themed we should update the list in README

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -429,6 +436,44 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_description_list(array &$va
...
+  }
+  $variables['items'] = [];
+
+  /** @var \Drupal\paragraphs\Entity\Paragraph $sub_paragraph */
+  foreach ($paragraph->get('field_oe_paragraphs')->referencedEntities() as $sub_paragraph) {
+    // Get the paragraph translation.
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    $description = '';
+    if (!$sub_paragraph->get('field_oe_title')->isEmpty()) {
```
drishu: is this a mistake of intended, what's the relationship between `field_oe_title` and `field_oe_plain_text_long` ?

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -429,6 +436,44 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_description_list(array &$va
...
+      $description = $sub_paragraph->get('field_oe_plain_text_long')->value;
+    }
+    $variables['items'][] = [
+      'icon' => str_replace('_', '-', $sub_paragraph->get('field_oe_icon')->value),
```
drishu: The values in IconOptionsSubscriber are with `-` already, did you encounter any issue ?

templates/paragraphs/paragraph--oe-facts-figures.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ * @file
+ * Theme override to display the 'Facts and Figures' #type paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{{ pattern('fact_figures', {
+  'title': title|default(''),
+  'items': items,
+  'columns': responsive_columns,
```
drishu: here is `responsive_columns` but in the preprocess I see `grid_columns`, and indeed when I test, I can set whichever number I like it will always present with 3

tests/src/Kernel/Paragraphs/FactsFiguresTest.php
```diff
@@ -0,0 +1,193 @@
...
+    $subtitle_fact = $crawler->filter('div.row-cols-md-3.row > div.col:nth-child(6) >  h5.fw-bold');
+    $this->assertStringContainsString(
+      'Figma components',
+      $subtitle_fact->html()
+    );
+    $description_fact = $crawler->filter('div.row-cols-md-3.row > div.col:nth-child(6) > p');
+    $this->assertStringContainsString(
+      'Sed efficitur bibendum rutrum. Nunc feugiat congue augue ac consectetur.',
+      $description_fact->html()
+    );
```
drishu, line 190: ok, we asserted the default layout in detail, but now, lets change to 1 column, then 2 columns and for each just assert that the general structure changed to match the columns saved

drishu (changes requested):

drishu (dismissed):

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -429,6 +436,44 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_description_list(array &$va
...
+      'subtitle' => $sub_paragraph->get('field_oe_subtitle')->value,
+      'description' => $description,
+    ];
+  }
+
+  if (!$paragraph->get('oe_bt_layout')->isEmpty()) {
+    $variables['columns'] = $paragraph->get('oe_bt_layout')->value;
+  }
+
+}
```
donquixote, line 476: Empty line at the end?

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -429,6 +436,44 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_description_list(array &$va
...
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_facts_figures(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value;
+  }
+
+  if (!$paragraph->get('field_oe_link')->isEmpty()) {
+    $variables['link_more']['path'] = $paragraph->get('field_oe_link')->first()->getUrl()->toString();
+    $variables['link_more']['label'] = $paragraph->get('field_oe_link')->first()->get('title')->getValue();
```
donquixote: Personal note:
We are calling `->get('field_oe_link')` 3 times.
We _could_ store the field item list or the field item from `->get('field_oe_link')->first()` in a local variable instead.
Would this give us anything other than a little bit of micro-optimization? Debatable.

The `->first()->get('title')->getValue()` could be replaced with just `->value`, I think.

I heard these points were already discussed with @drishu and @brummbar so let's keep it this way.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -429,6 +436,44 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_description_list(array &$va
   }
 }
 
+/**
+ * Implements hook_preprocess_paragraph().
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_facts_figures(array &$variables): void {
+  $paragraph = $variables['paragraph'];
```
donquixote, line 444: Can we put inline `/** @var \Drupal\paragraphs\ParagraphInterface $paragraph */` above this variable?

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -356,7 +360,10 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
  */
 function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables): void {
   $paragraph = $variables['paragraph'];
-  $variables['heading'] = $paragraph->get('field_oe_title')->value ?? '';
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['heading'] = $paragraph->get('field_oe_title')->value;
+  }
```
donquixote, line 365: Personal note:
It seems to me that technically, the `->isEmpty()` is not needed in this case. `->value` (via magic `__get()`) will simply give you `NULL` if the field is empty. (unlike `->first()->*` which gives trouble if `->first()` returns NULL.)
(An edge case could be if the database contains leftover empty-ish field items, where `->isEmpty()` would return `TRUE` but `->value` would give you something else. I don't know if this can ever happen, and if it is relevant for most of the stuff we do.)

Is it more semantic if we call `->isEmpty()`? Maybe.
To me it feels like unnecessary verbosity and repetition, but that's just me.

I heard that @brummbar and @drishu discussed this and decided that we should always call `->isEmpty()` if the field is optional - or really always? If that's their final word then ¯\_(ツ)_/¯.

This note would apply to more cases than just here.

donquixote: Perhaps the idea for us using `->isEmpty()` is to clearly identify optional fields.
I don't have a clear opinion on it but it would be a valid reason.

templates/patterns/fact_figures/fact_figures.ui_patterns.yml
```diff
@@ -0,0 +1,51 @@
...
+      escape: false
+      preview: 2
+    items:
+      type: array
+      label: 'Items'
+      description: 'List of facts, with icon, title, subtitle and content.'
+      preview:
+        - icon: 'x-diamond-fill'
+          subtitle: 'Jira Tickets'
+          title: '1529 JIRA Ticket'
```
donquixote, line 22: Shouldn't we put the `title` above the `subtitle` in these preview items?

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+    return [
+      IconOptionsEvent::class => ['getIconOptions', -1],
+    ];
+  }
+
+  /**
+   * Gets the icon options.
+   *
+   * @param \Drupal\custom_events\Event\IconOptionsEvent $event
+   *   Allowed format event object.
```
donquixote, line 31: The namespace for the type in the `@param` doc is wrong. It should be `\Drupal\oe_paragraphs\Event\IconOptionsEvent`.

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+declare(strict_types = 1);
+
+namespace Drupal\oe_bootstrap_theme_paragraphs\EventSubscriber;
+
+use Drupal\oe_paragraphs\Event\IconOptionsEvent;
+use Symfony\Component\EventDispatcher\EventSubscriberInterface;
+
+/**
+ * Provides options for the icon field.
+ */
```
donquixote, line 15: I have questions about this mechanism, which I would like to see explained in the class doc or perhaps in a method doc.

1. Which icon field does this target, and how? From looking at other code I see this is targeting `field_oe_icon` which is defiend in oe_paragraphs. A `@see _oe_paragraphs_allowed_values_icons()` doc could help.
2. Which functionality from oe_paragraphs are we overriding here? Perhaps a `@see \Drupal\oe_paragraphs\EventSubscriber\OptionsSubscriber` doc can help.
3. Where does the list of icons come from? I assume we have an icons library somewhere in BCL? How do the machine names map to real icons? Is this a full list of icons available in bcl, or just a subset?

tests/src/Functional/ParagraphsTest.php
```diff
@@ -140,7 +140,83 @@ public function testAccordionParagraph(): void {
...
+    ];
+    foreach ($allowed_values as $allowed_value) {
+      $this->assertSession()->elementsCount('css', 'option[value="' . $allowed_value . '"]', 1);
+    }
+    $this->assertSession()->elementsCount('css', 'select#edit-oe-bt-paragraphs-0-subform-field-oe-icon option', 20);
+
+  }
+
+  /**
+   * Adds a field to a paragraph.
```
donquixote: What happened to this doc comment? It used to be "Test Description list paragraphs form." which seems more correct than the new doc comment.

tests/src/Kernel/Paragraphs/FactsFiguresTest.php
```diff
@@ -0,0 +1,216 @@
...
+    $paragraph->save();
+
+    // Testing: Facts and figures - Default layout.
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    ini_set('xdebug.var_display_max_depth', '10');
+    ini_set('xdebug.var_display_max_children', '256');
+    ini_set('xdebug.var_display_max_data', '100024');
+    var_dump($crawler->filter('body')->html());
```
donquixote: Is the ini_set() and the var_dump() leftover code, or is there a reason for it?
If it is needed for some reason, I would expect a comment with an explanation.

modules/oe_bootstrap_theme_paragraphs/config/post_updates/00003/field.storage.paragraph.oe_bt_layout.yml
```diff
@@ -0,0 +1,19 @@
+langcode: en
+status: true
+dependencies:
+  module:
+    - paragraphs
+id: paragraph.oe_bt_layout
+field_name: oe_bt_layout
```
donquixote: I am not happy with the field name.
In the future we might want to create other fields named `oe_bt_layout` which would not be integer fields, but a select similar to the "variant". To me "layout" does not imply an integer field.
Alternative would be "oe_bt_n_columns".

But: If this was already agreed on with @drishu and @brummbar then I accept it.

modules/oe_bootstrap_theme_paragraphs/config/post_updates/00003/field.field.paragraph.oe_facts_figures.oe_bt_layout.yml
```diff
@@ -0,0 +1,24 @@
...
+status: true
+dependencies:
+  config:
+    - field.storage.paragraph.oe_bt_layout
+    - paragraphs.paragraphs_type.oe_facts_figures
+id: paragraph.oe_facts_figures.oe_bt_layout
+field_name: oe_bt_layout
+entity_type: paragraph
+bundle: oe_facts_figures
+label: 'Layout'
```
donquixote: In the edit form I would expect a label like "Number of columns", not "Layout".
But if this was already decided by others, then what can I say..

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsSubscriber.php
```diff
@@ -0,0 +1,54 @@
...
+      'file' => 'File',
+      'image' => 'Image',
+      'info' => 'Info',
+      'linkedin' => 'LinkedIn',
+      'files' => 'Multiple files',
+      'rss' => 'RSS',
+      'search' => 'Search',
+      'share' => 'Share',
+      'twitter' => 'Twitter',
+      'person-video' => 'Video',
```
donquixote: For some reason, "Video" icon does not work for me locally.
Are we sure that all the icons exist in BCL?

donquixote (changes requested):
I left some notes. For each of them you can either:
- Change as suggested, or
- Explain that this was already discussed and should stay the way it is.

donquixote:
I am updating via merge, to not rip the context from review comments.
We can still rebase it later, when we are happy with all changes.

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsSubscriber.php
abel-santos-corral: Right, now we don't have the person-video on file bootstrap-icons.svg. At the time when we did it it was there. So we must check if we can add that icon now. @escuriola we must check now at the file what we have available. At the time I did this, I checked at the bootstrap icon site, because the FRONT team told us that the all icons from [here](https://icons.getbootstrap.com/) were on the file.

modules/oe_bootstrap_theme_paragraphs/config/post_updates/00003/field.field.paragraph.oe_facts_figures.oe_bt_layout.yml
abel-santos-corral: See comment above.

modules/oe_bootstrap_theme_paragraphs/config/post_updates/00003/field.storage.paragraph.oe_bt_layout.yml
abel-santos-corral: In fact this field was formerly the 'variant' field. https://github.com/openeuropa/oe_bootstrap_theme/pull/123/commits/41a02e6d538438fdd749bff4d8a746ac963b461e

To check if is worth the time of changing the name again.

Another consideration we did at the time was to put oe_bt_<name-field> and try to make the field reusable transversally through all paragraphs. So if this oe_bt_n_columns is a better and more reusable field, then let's use it.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
escuriola: The title field value can not be replaced with ->value in this case.

modules/oe_bootstrap_theme_paragraphs/src/EventSubscriber/IconOptionsSubscriber.php
escuriola: 3. Icons subset from BCL

donquixote: > 3. Icons subset from BCL
I see the icons in bootstrap-icons.svg.
I still don't see a doc in the code explaining this.. But ok.

modules/oe_bootstrap_theme_paragraphs/config/post_updates/00003/field.storage.paragraph.oe_bt_layout.yml
donquixote: Conclusion from @brummbar : Yes we use oe_bt_n_columns. And label can be "Number of columns".
(There was an idea to already use oe_w_* prefix in inticipation of the move, but imo it is cleaner to stick with oe_bt_* while the code is here)

drishu (changes requested):
implement the "oe_bt_n_columns" as requested by @donquixote, then I give it another functional test and we are good to go

drishu (dismissed):

donquixote (dismissed):
I am happy now :)

drishu (approved):

donquixote (approved):

donquixote:
Old review comment by Carlos:

> Missing also a FactsAndFiguresTest.php in test > src > Kernel > Paragraphs

This seems to be fixed :)

PR #127, opened 2021-11-29
OEL-494: Create 'File' paragraph component

Maxfire:
Jira issue(s):
- [OEL-494](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-494)


GilNovacomm (dismissed):
All is well.

templates/patterns/file/pattern-file.html.twig
```diff
@@ -0,0 +1,29 @@
+{#
+/**
+ * @file
+ * File download pattern.
+ */
+#}
+{%  if file is not empty %}
+  {% set bcl_icon_path = '/assets/icons/bootstrap-icons.svg' %}
```
brummbar: Let's do like this https://github.com/openeuropa/oe_theme/blob/3.x/oe_theme.theme#L74 .
We are introducing this path over and over, but we can just put a generic preprocess and defined the value there.
Search for the path and remove it from everywhere in this PR itself.

templates/patterns/file/pattern-file.html.twig
```diff
@@ -0,0 +1,29 @@
...
+    },
+    'download': {
+      'label': button_label|default('Download'|t),
+      'path': file.url,
+      'icon': {
+        'name': 'download',
+        'size': 'fluid',
+      },
+    },
+    translation: translation,
```
brummbar: Why we are mixing quoted and unquoted?

templates/patterns/file/file.ui_patterns.yml
```diff
@@ -0,0 +1,43 @@
...
+      description: "A file value object."
+      preview:
+        title: "Example document"
+        extension: "pdf"
+        url: "http://example.com/document.pdf"
+        size: "4444"
+        name: "document.pdf"
+        languageCode: "Français"
+    translation:
+      type: object
```
brummbar: Why we don't use array here?

templates/patterns/file/file.ui_patterns.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: "Translation"
+      escape: false
+      preview:
+        id: ""
```
brummbar: This shouldn't be needed.

templates/patterns/file/file.ui_patterns.yml
```diff
@@ -0,0 +1,43 @@
...
+        label:
+          label: "Other languages (3)"
+          icon_position: "after"
```
brummbar: Let's simplify this by removing icon position. We pass a string only to label.

tests/src/Kernel/Paragraphs/FileParagraphTest.php
```diff
@@ -0,0 +1,155 @@
...
+   * Test file pattern rendering.
+   *
+   * @param array $file
+   *   A file array.
+   *
+   * @throws \Exception
+   *
+   * @dataProvider dataProvider
+   */
+  public function testFilePatternRendering(array $file) {
```
brummbar: We shouldn't test the pattern here, for that we have the markup rendering tests. We need to test only the paragraph.

tests/src/Kernel/Paragraphs/FileParagraphTest.php
```diff
@@ -0,0 +1,155 @@
...
+use Drupal\Core\Site\Settings;
+use Drupal\file\Entity\File;
+use Drupal\oe_bootstrap_theme\ValueObject\FileValueObject;
+use Drupal\Tests\oe_bootstrap_theme\Kernel\Paragraphs\ParagraphsTestBase;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Test file pattern rendering.
+ */
+class FileParagraphTest extends ParagraphsTestBase {
```
brummbar, line 13: We are overriding the form above, so we need to test the paragraph creation too.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -311,6 +313,87 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
...
+  if (isset($translation_file_value)) {
+    $variables['translation']['id'] = 'language-dropdown' . rand(0, 99);
+    $variables['translation']['label'] = [
+      'label' => t('Other languages (@languages)', ['@languages' => $num_languages]),
+      'icon position' => 'after',
+      'icon' => [
+        'name' => 'caret-down-fill',
+        'size' => '2xs',
+      ],
+    ];
```
brummbar: This is not correct, it should be all handled in the template and not at paragraph level.

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_form_display.media.document.default.yml
```diff
@@ -0,0 +1,69 @@
+langcode: en
```
brummbar: Why is this overridden, I didn't find significant differences with the one provided by oe_media.

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_form_display.paragraph.oe_document.default.yml
```diff
@@ -0,0 +1,29 @@
+langcode: en
```
brummbar: Same in here, the only extra is the translation field, which shouldn't be exported (make sure to not have multilingual modules enabled when exporting the configuration).

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_view_display.media.document.default.yml
```diff
@@ -0,0 +1,45 @@
+langcode: en
```
brummbar: Same with this file, looks the same as oe_paragraphs to me.

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_view_display.paragraph.oe_document.default.yml
```diff
@@ -0,0 +1,22 @@
+langcode: en
```
brummbar: Same.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -311,6 +313,87 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
...
+        'download' => [
+          'label' => t('Download'),
+          'path' => $translation_file_value->getUrl(),
+        ],
+      ];
+    }
+    $num_languages++;
+  }
+  if (isset($translation_file_value)) {
+    $variables['translation']['id'] = 'language-dropdown' . rand(0, 99);
```
brummbar: If we have to generate an ID, always use `\Drupal\Component\Utility\Html::getUniqueId()`.

src/DocumentMediaValueExtractor.php
```diff
@@ -0,0 +1,86 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\oe_bootstrap_theme;
+
+use Drupal\media\MediaInterface;
+use Drupal\oe_bootstrap_theme\ValueObject\FileValueObject;
+
+/**
+ * Extracts a FileValueObject from a document media.
```
brummbar, line 11: Put a comment that this is extracted from oe_theme.

brummbar: Let's also add `@internal` for now.

modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.services.yml
```diff
@@ -0,0 +1,5 @@
+services:
+  oe_bootstrap_theme_helper.twig_extension.twig_extension:
```
brummbar:
```suggestion
  oe_bootstrap_theme_helper.twig_extension:
```

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -0,0 +1,88 @@
...
+      'file-pdf-fill' => [
+        'pdf',
+        'application/pdf',
+      ],
+      'file-word-fill' => [
+        'word',
+      ],
+    ];
+
+    foreach ($extension_mapping as $file_type => $extensions) {
```
brummbar:
```suggestion
    foreach ($extension_mapping as $icon => $extensions) {
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -311,6 +313,87 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
...
+/**
+ * Implements hook_preprocess_paragraph() for oe_document paragraph.
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_document(array &$variables): void {
+  $media = $variables['content']['field_oe_media'][0]['#media'];
+  if (!isset($media)) {
+    return;
+  }
+
+  $type = $media->get('oe_media_file_type')->value;
```
brummbar: Most of this code is done inside `DocumentMediaValueExtractor::getFileValue()`, the way this is implemented now I get a critical error due to missing extension when I upload a remote file.
Let's follow a bit more how oe_theme is done, as I think it's beneficial:

1. Let's style the media entity itself, instead of the paragraph. Move to media entity preprocess.
2. The paragraph formatter will simply render the entity with the default view mode (look in `oe_theme_preprocess_media__document__default()`).

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -311,6 +313,87 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
...
+  if ($type == "local") {
+    $file_entity = File::load($media->get('oe_media_file')->target_id);
+    if (!isset($file_entity)) {
+      return;
+    }
+
+    // File operations.
+    $file_value = DocumentMediaValueExtractor::getFileValue($media);
+    $variables['file'] = $file_value;
+    $variables['file']->setLanguageCode($file_entity->language()->getName());
```
brummbar: The language code isn't a name, it's the code itself.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -311,6 +313,87 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_banner(array &$variables):
...
+      $language_config = $language_config_storage->load($translation_file_value->getLanguageCode());
+      $variables['translation']['items'][] = [
+        'language' => $language_config->label(),
+        'meta' => '(' . format_size($translation_file_value->getSize()) .
+        ' - ' . strtoupper($translation_file_value->getExtension()) . ')',
+        'download' => [
+          'label' => t('Download'),
+          'path' => $translation_file_value->getUrl(),
+        ],
+      ];
```
brummbar: All the meta preprocessing should go in preprocess functions if we use a single meta field. Check oe_theme, we have checks for missing file size too.

src/ValueObject/FileValueObject.php
```diff
@@ -0,0 +1,267 @@
...
+  /**
+   * Getter.
+   *
+   * @return string
+   *   Property value.
+   */
+  public function getExtension(): string {
+    $mime_types = new MimeTypes();
+    $extension = $mime_types->getExtension($this->getMime());
+    return $extension ?? pathinfo($this->name, PATHINFO_EXTENSION);
```
brummbar: Let's improve on oe_theme by doing this:
```
    if ($this->extension === NULL) {
      $mime_types = new MimeTypes();
      $extension = $mime_types->getExtension($this->getMime());
      $this->extension = $extension ?? pathinfo($this->name, PATHINFO_EXTENSION);
    }

    return $this->extension;
```

Note: I have this code because I was debugging one error reported above and I kept getting inside this function over and over and over... 😄

templates/patterns/file/file.ui_patterns.yml
```diff
@@ -0,0 +1,43 @@
...
+      label: "File"
+      description: "A file value object."
+      preview:
+        title: "Example document"
+        extension: "pdf"
+        url: "http://example.com/document.pdf"
+        size: "4444"
+        name: "document.pdf"
+        languageCode: "Français"
+    translation:
```
brummbar: This can be multiple translations, so let's use translation**s**.

templates/patterns/file/file.ui_patterns.yml
```diff
@@ -0,0 +1,43 @@
...
+        id: ""
+        label:
+          label: "Other languages (3)"
+          icon_position: "after"
+        items:
+          - language: "български"
+            meta: "(16.2 MB - PDF)"
+            download:
+              label: "Download"
+              path: "/example.html"
```
brummbar: Let's simplify for the download label, we don't need to specify it in ALL the translations. Let's put a button_label on the top level, like it's done in oe_theme. So we avoid to repeat it here.
Fine to keep meta as single entry, but then we need to pass the processed meta information from the preprocess.
Download becomes url with a sigle path to the file, language we put the language_code and translate when passing down to the bcl template.

brummbar (changes requested):

templates/patterns/file/pattern-file.html.twig
escuriola: We already have this in oe_bootstrap_theme.theme (https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/oe_bootstrap_theme.theme#L20)

templates/patterns/file/file.ui_patterns.yml
escuriola: Not compatible with bcl-file component

modules/oe_bootstrap_theme_paragraphs/config/overrides/core.entity_form_display.media.document.default.yml
escuriola: None of these files were replaced because they were not listed in oe_bootstrap_theme_paragraphs.install

templates/patterns/file/pattern-file.html.twig
brummbar: Then why we keep defining the variable over and over and over?

composer.json
```diff
@@ -22,7 +23,8 @@
         "openeuropa/code-review": "~1.6",
         "openeuropa/drupal-core-require-dev": "^8.9 || ^9.1",
         "openeuropa/oe_media": "^1.14",
-        "openeuropa/oe_paragraphs": "^1.10",
+        "openeuropa/oe_multilingual": "^1.8",
+        "openeuropa/oe_paragraphs": "dev-master",
```
escuriola: After the meeting with Francesco about testing in oe_showcase I think oe_paragraph, oe_media, etc., must be moved to require section but please confirm with Francesco.

escuriola (changes requested):

escuriola (dismissed):

PR #131, opened 2021-12-01
OEL-601: Submit button template modified to add button pattern.

Maxfire:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-601


Maxfire:
Closing since it's not needed.

PR #136, opened 2021-12-07
OEL-837: Fixes to upgrade BCL to v0.15.0

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-837


escuriola:
We already upgrade to BCL 0.19.0

PR #137, opened 2021-12-07
OEL-753: Integrate inpage navigation

abel-santos-corral:
Jira issue(s):
- [OEL-753](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-753)

(!) Don't merge until OEL-824 is merged at 1.x.


kits/default/package.json
```diff
@@ -13,7 +13,7 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "0.14.0",
+    "@openeuropa/bcl-builder": "0.15.0",
```
escuriola: This version is not needed to this task and blocks the deploy, we can back to the 14.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
 
+/**
+ * Implements hook_preprocess_paragraph() for paragraph--oe-content-row--variant-inpage-navigation.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_content_row__variant_inpage_navigation(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
+  }
```
escuriola: The if is not needed
```suggestion
  $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+
+  $variables['full_layout'] = TRUE;
+  $variables['scrollspy_id'] = 'bcl-inpage-navigation';
```
escuriola: Concatenate with a random number/string to avoid problems if more than one scrollspy are displayed at the same page.

package.json
```diff
@@ -13,8 +13,8 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "0.14.0",
-    "@openeuropa/bcl-theme-default": "0.14.0",
+    "@openeuropa/bcl-builder": "0.15.0",
+    "@openeuropa/bcl-theme-default": "0.15.0",
```
escuriola: Same as above, drop this change.

resources/js/libraries/extend/inpage_navigation.js
```diff
@@ -0,0 +1,9 @@
+(function (bs, Drupal) {
+  Drupal.behaviors.inpage_navigation = {
+    attach: function attach(context, settings) {
+      new bs.ScrollSpy(document.body, {
+        target: '#bcl-inpage-navigation'
```
escuriola, line 6: Same as above, what about we have more than one scrollspy in the same page? Maybe the target must be a class and not an id, we can talk about it with someone else...

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -0,0 +1,15 @@
...
+#}
+{{ attach_library('oe_bootstrap_theme/drupal.inpage_navigation') }}
+
+{% include '@oe-bcl/bcl-inpage-navigation/inpage-navigation.html.twig' with {
+  'title': title,
+  'full_layout': full_layout,
+  'links': links,
+  'content': content,
+  'id': scrollspy_id,
+} %}
```
escuriola: Missing only and attributes

escuriola (changes requested):

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
abel-santos-corral: Not possible until delivery of BCL Patterns 0.16.0 where Front team delivers the fix of the OEL-875. Now only one inpage navigation can be created and must have that id to include top and background-color.

resources/js/libraries/extend/inpage_navigation.js
abel-santos-corral: We need this until BCL 16 is delivered. Then we will change this here and use the class.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,64 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
   $variables['hide_from'] = $paragraph->get('field_oe_timeline_expand')->value;
 }
 
+/**
+ * Implements hook_preprocess_paragraph() for paragraph--oe-content-row--variant-inpage-navigation.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_content_row__variant_inpage_navigation(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+
+  $variables['title'] = $paragraph->get('field_oe_title')->value ?? '';
```
escuriola: Remember this morning, we have to change this to isEmpty()

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,64 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    elseif ($sub_paragraph->hasField('field_oe_text') && !$sub_paragraph->get('field_oe_text')->isEmpty()) {
+      $links_label = $sub_paragraph->get('field_oe_text')->first()->value;
+    }
+    elseif ($sub_paragraph->hasField('field_oe_plain_text_long') && !$sub_paragraph->get('field_oe_plain_text_long')->isEmpty()) {
+      $links_label = $sub_paragraph->get('field_oe_plain_text_long')->first()->value;
+    }
+    elseif ($sub_paragraph->bundle() == 'oe_accordion') {
+      $accordion_items = $sub_paragraph->get('field_oe_paragraphs')->referencedEntities();
+      $first_accordion = $accordion_items[0];
+      $links_label = $first_accordion->get('field_oe_text')->first()->value;
```
escuriola: isEmpty?

escuriola (changes requested):

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
abel-santos-corral: This one is a required field of OE Accordion item (you can check at http://localhost:8080/build/admin/structure/paragraphs_type/oe_accordion_item/fields/paragraph.oe_accordion_item.field_oe_text_long). So I didn't use the isEmpty.

escuriola (dismissed):

resources/js/libraries/extend/inpage_navigation.js
drishu: agreed but add a todo referring to a ticket or issue

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+ * Implements hook_preprocess_paragraph() for paragraph--oe-content-row--variant-inpage-navigation.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_content_row__variant_inpage_navigation(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value;
+  }
+
+  $variables['full_layout'] = TRUE;
```
drishu: I don't see this in bcl

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+  $variables['scrollspy_id'] = 'bcl-inpage-navigation';
```
drishu: in case this id is mandatory please add a todo refering to a follow up ticket that fixes this.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+      $links_label = $sub_paragraph->get('field_oe_plain_text_long')->first()->value;
+    }
+    elseif ($sub_paragraph->bundle() == 'oe_accordion') {
+      $accordion_items = $sub_paragraph->get('field_oe_paragraphs')->referencedEntities();
+      $first_accordion = $accordion_items[0];
+      $links_label = $first_accordion->get('field_oe_text')->first()->value;
+    }
+
+    $unique_id = Html::getUniqueId('bcl-inpage-item-' . $sub_paragraph->id());
+    // Get sub-paragraph translation.
```
drishu:
```suggestion
    // Get paragraph translation.
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    $entity_type = 'paragraph';
+    $view_mode = 'default';
```
drishu: $view_mode is used only once, no point in holding the memory, u can pass it as a static string

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+  $links = [];
+  $content = [];
+
+  foreach ($paragraph->get('field_oe_paragraphs')->referencedEntities() as $delta => $sub_paragraph) {
+    /** @var \Drupal\paragraphs\Entity\Paragraph $sub_paragraph */
+    if ($sub_paragraph->hasField('field_oe_title') && !$sub_paragraph->get('field_oe_title')->isEmpty()) {
+      $links_label = $sub_paragraph->get('field_oe_title')->first()->value;
+    }
+    elseif ($sub_paragraph->hasField('field_oe_text') && !$sub_paragraph->get('field_oe_text')->isEmpty()) {
+      $links_label = $sub_paragraph->get('field_oe_text')->first()->value;
```
drishu: you have to explain this one to me, why do we use these backups ?

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,66 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    }
+    elseif ($sub_paragraph->hasField('field_oe_plain_text_long') && !$sub_paragraph->get('field_oe_plain_text_long')->isEmpty()) {
+      $links_label = $sub_paragraph->get('field_oe_plain_text_long')->first()->value;
```
drishu: this one looks scary, how long can this be ?

templates/paragraphs/paragraph--oe-content-row--variant-inpage-navigation.html.twig
```diff
@@ -0,0 +1,15 @@
+{#
+/**
+ * @file
+ * Theme override to display the 'Content row' #type and 'Inpage navigation' #variant paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{{ pattern('inpage_navigation', {
+  'title': title|default(''),
```
drishu: The field can be empty so I would not add a default like in oe_theme, we don't have those requirements, that's good, but, what's the point of setting an empty default?
```suggestion
  'title': title,
```

templates/paragraphs/paragraph--oe-content-row--variant-inpage-navigation.html.twig
```diff
@@ -0,0 +1,15 @@
...
+  'full_layout': full_layout,
```
drishu: remove this, I don't see this in bcl

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,37 @@
+inpage_navigation:
+  label: "Inpage navigation"
+  description: "Inpage navigation adds a navigation element and content element so when scrolling the active element is indicated in the nav part."
+  settings:
+    full_layout:
```
drishu: we are at version 14 of bcl in this branch, this is not part of that and I can't see the effect of adding and using this, please remove this until it is available, the upgrade ticket need to handle this

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,37 @@
...
...n pariatur ex labore qui laborum sint consequat. Aute cupidatat id sint incididunt Lorem Lorem aute aliqua sint aliquip commodo laborum quis. Ex irure sit reprehenderit in amet. Fugiat in amet proident consequat occaecat officia et consequat ipsum aute commodo. Pariatur duis consequat ipsum enim quis Lorem sint eiusmod esse. Occaecat amet esse aliquip non incididunt nisi sint do. Magna deserunt irure est nostrud deserunt in elit in enim do occaecat consequat. Incididunt duis reprehenderit commodo mollit culpa ex. Cillum ullamco cupidatat sit fugiat exercitation voluptate elit nulla duis. Dolor deserunt ad aute sint. Sunt cillum aute anim quis anim anim. Voluptate enim proident voluptate deserunt mollit. Veniam exercitation ex aliquip laborum commodo esse laboris. Veniam exercitation dolore quis ea minim magna. Labore adipisicing magna in excepteur esse ullamco aliqua tempor veniam id anim. Nulla deserunt laboris aute reprehenderit nisi quis commodo. Excepteur tempor nulla amet fugiat ullamco sit. Nostrud proident exercitation veniam elit ullamco ullamco officia minim officia commodo id duis. Laboris ullamco aliquip voluptate dolore aliquip nostrud magna ut cillum nisi officia.</p>"
```
drishu: can we make the content a bit smaller, just seems a bit much

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -0,0 +1,16 @@
...
+/**
+ * @file
+ * Inpage navigation pattern.
+ */
+#}
+{{ attach_library('oe_bootstrap_theme/drupal.inpage_navigation') }}
+
+{% include '@oe-bcl/bcl-inpage-navigation/inpage-navigation.html.twig' with {
+  'title': title,
+  'full_layout': full_layout,
```
drishu: remove this until avail

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -0,0 +1,16 @@
...
+  'links': links,
+  'content': content,
+  'id': scrollspy_id,
+  'attributes': attributes,
```
drishu: we should add this only if it fixes something (and even so use create_attribute), but I don't think we need to be fixing anything here

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
```diff
@@ -0,0 +1,93 @@
...
+        ]),
+        4 => Paragraph::create([
+          'type' => 'oe_rich_text',
+          'field_oe_title' => 'Title rich text example 5',
+          'field_oe_text_long' => 'Curabitur efficitur ultrices consectetur. Nulla bibendum gravida quam a aliquet. Duis non accumsan enim. Phasellus tincidunt venenatis magna, at commodo nisi. Vivamus nec nibh vestibulum, pellentesque dolor vel, vehicula sem. Mauris lacus nulla, interdum eu justo id, sagittis tempor nulla. Phasellus faucibus, libero congue blandit sagittis, sem augue commodo enim, a tristique massa lorem vel justo. Nunc pretium consectetur felis sed consequat. Vivamus ac rutrum magna, sed laoreet turpis. Sed nisi lacus, volutpat eget mi id, maximus congue quam. Cras dictum a odio nec convallis. Mauris tincidunt felis ut eros auctor, ut varius augue rutrum. Pellentesque habitant morbi tristique senectus et netus et malesuada fames ac turpis egestas. Pellentesque posuere ornare augue. Integer ornare sagittis risus et viverra. Sed eu nulla porta, efficitur ante id, tempus nunc.',
+        ]),
+        5 => Paragraph::create([
+          'type' => 'oe_rich_text',
```
drishu: I don't see the point in adding so many reduce the number of them

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
```diff
@@ -0,0 +1,93 @@
...
+    $this->assertCount(1, $crawler->filter('div.col-md-3.d-none.d-md-block'));
+    $this->assertCount(1, $crawler->filter('nav.position-sticky'));
+    $this->assertCount(1, $crawler->filter('h5'));
+    $this->assertCount(1, $crawler->filter('ul.nav.nav-pills.flex-column'));
+    $this->assertCount(6, $crawler->filter('li.nav-item'));
+    $this->assertCount(6, $crawler->filter('a.nav-link'));
+    $this->assertCount(1, $crawler->filter('div.col-md-9'));
+    $this->assertCount(6, $crawler->filter('p'));
+
+    for ($i = 1; $i <= 6; $i++) {
```
drishu: for loops in test have the tendency of masking failing scenarios, not necessarily the case here but I would not make a habit out it, let's have less items to test and list them one by one instead of a loop

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
```diff
@@ -0,0 +1,93 @@
...
+    // Create sub-paragraphs for OE Content row paragraph.
+    $sub_paragraphs =
+      [
+        0 => Paragraph::create([
+          'type' => 'oe_rich_text',
+          'field_oe_title' => 'Title rich text example 1',
+          'field_oe_text_long' => 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Pellentesque ornare et elit a dictum. Maecenas lacinia eros quis eros iaculis, sit amet bibendum massa facilisis. Integer arcu nisl, fringilla nec quam vel, tincidunt maximus ex. Suspendisse ac arcu efficitur, feugiat tellus vel, viverra sapien. Etiam vitae condimentum lorem. Nulla congue ligula lacinia efficitur tempus. Duis vitae auctor enim. Nulla iaculis, diam et sagittis scelerisque, est mauris luctus sem, a imperdiet lacus diam eu dui. Morbi accumsan, augue eu gravida elementum, libero mi blandit odio, eu fringilla nunc ipsum non tellus. Suspendisse dapibus elit at lobortis pretium. Quisque vestibulum ut purus sit amet molestie. Sed eget volutpat justo, vel varius augue. Vestibulum vel risus facilisis, feugiat sem aliquam, lobortis ante.',
+        ]),
+        1 => Paragraph::create([
+          'type' => 'oe_rich_text',
```
drishu: add also other types

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -0,0 +1,77 @@
+inpage_navigation_full:
+  render:
+    '#type': pattern
+    '#id': inpage_navigation
+    '#fields':
+      settings:
+        full_layout: true
```
drishu: update this test one you have changes above done

drishu (changes requested):
Please also check drone before you send this to review!

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
abel-santos-corral: This are no backups, we manage three kind of paragraphs that can be added to a OE content row paragraph:
* Paragraphs which include a field called field_oe_title, then the title must be used at the inpage menu at the left part of the Content Row, and set an anchor for each title.
* Paragraphs which include a field field_oe_text which is used as a title for that paragraph type. Then we use that to mount the inpage navigation.
* Paragraph accordion made of n paragraph accordion items. Paragraph accordion per se doesn't have title, so we must use the title of the first Paragraph accordion item to use it to mount the navigation.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
abel-santos-corral: We need to check the paragraph type Accordion. If we place an accordion in the inpage navigation we need to mount also the inpage stuff. 

Please, check it on a live case to understand what problems we have here. Then we can discuss the solutions (E.g: a) Add a title to an Accordion paragraph, b) Add code to avoid Accordion to be included in an inpage navigation....).

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
abel-santos-corral: In order to see the effect of inpage we need a longer text than in the rest of elements, which you can see at /patterns page or at the example of netlify. 
If you still need a shorter text, I'll put it, but consider that then the effect of scrollspy can be lost if text is small.

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
abel-santos-corral: We, we must provide that an Inpage navigation is mounted for all and each of the paragraphs that can be added. Why? Simply because there's code at .module file that must be tested and ensure that all elements are shown correctly. With manual testing of this can be better understood.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
abel-santos-corral: This setting is only present at pattern side, not at paragraph side. So removed from here and set as TRUE directly in the twig of the paragraph. Check that you agree on this change.

templates/paragraphs/paragraph--oe-content-row--variant-inpage-navigation.html.twig
abel-santos-corral: Right, in v.0.14.0 this is not there. We started with 14 --> passed to 15 and then back to 14. So I'll remove this stuff.

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
abel-santos-corral: OK, removing it, but please read comment: https://github.com/openeuropa/oe_bootstrap_theme/pull/137#discussion_r765048075. @drishu check it and then let's clarify with team. As per isEmpty: same criteria must be applied at all reviews.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,65 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+ */
+function oe_bootstrap_theme_preprocess_paragraph__oe_content_row__variant_inpage_navigation(array &$variables): void {
+  $paragraph = $variables['paragraph'];
+
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value;
+  }
+
+  // @todo https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-848 will fix this to use class.
+  $variables['scrollspy_id'] = 'bcl-inpage-navigation';
```
drishu: thinking about this I think it doesn't make sense exposing it, let's move this into the pattern template with todo and all

resources/js/libraries/extend/inpage_navigation.js
```diff
@@ -0,0 +1,10 @@
+(function (bs, Drupal) {
+  Drupal.behaviors.inpage_navigation = {
+    attach: function attach(context, settings) {
+      new bs.ScrollSpy(document.body, {
+        // TODO: At ticket OEL-848 this will be fixed to target on class and not on id.
+        target: '#bcl-inpage-navigation'
```
drishu, line 6: not blocking now, but we have to think about this, once we have the unique ids how will we know the id to target?

templates/paragraphs/paragraph--oe-content-row--variant-inpage-navigation.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ * Theme override to display the 'Content row' #type and 'Inpage navigation' #variant paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{{ pattern('inpage_navigation', {
+  'title': title,
+  'links': links,
+  'content': content,
+  'scrollspy_id': scrollspy_id,
```
drishu: remove this, we move it to pattern template only

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,32 @@
+inpage_navigation:
+  label: "Inpage navigation"
+  description: "Inpage navigation adds a navigation element and content element so when scrolling the active element is indicated in the nav part."
+  settings:
+    scrollspy_id:
```
drishu: remove

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,65 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+
+  $links = [];
+  $content = [];
+
+  foreach ($paragraph->get('field_oe_paragraphs')->referencedEntities() as $delta => $sub_paragraph) {
```
drishu, line 351: I have error "Notice: Undefined variable: links_label in oe_bootstrap_theme_preprocess_paragraph__oe_content_row__variant_inpage_navigation() (line 384"
we should check for some empty values and exit if needed, like here https://github.com/openeuropa/oe_theme/blob/3.x/oe_theme.theme#L890

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
drishu: Now I understand, code like this always points to a bigger issue, and in this case it's the fact that you need a title else you don't print the paragraph into the page.
You should not be overwriting $variables['content']

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
drishu: Normally you should check first since you wrote the code, I did and indeed it can be way to long, this is a bad idea, we should use just field_oe_title

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,65 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    $content[] = $current_paragraph;
+
+    $links[] = [
+      'path' => '#' . $unique_id,
+      'label' => $links_label,
+    ];
+  }
+
+  $variables['links'] = $links;
+  $variables['content'] = $content;
```
drishu: remove

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,65 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+
+    $view_builder = \Drupal::entityTypeManager()->getViewBuilder($entity_type);
+    $storage = \Drupal::entityTypeManager()->getStorage($entity_type);
+    $paragraph = $storage->load($sub_paragraph->id());
+    $current_paragraph = $view_builder->view($paragraph, 'default');
+    // Wrap the paragraph in a div with a specific id set as anchor.
+    $current_paragraph['#theme_wrappers']['container'] = [
+      '#attributes' => ['id' => $unique_id],
+    ];
+    $content[] = $current_paragraph;
```
drishu: remove

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,65 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    $sub_paragraph = \Drupal::service('entity.repository')
+      ->getTranslationFromContext($sub_paragraph, $paragraph->language()->getId());
+    $entity_type = 'paragraph';
+
+    $view_builder = \Drupal::entityTypeManager()->getViewBuilder($entity_type);
+    $storage = \Drupal::entityTypeManager()->getStorage($entity_type);
+    $paragraph = $storage->load($sub_paragraph->id());
+    $current_paragraph = $view_builder->view($paragraph, 'default');
+    // Wrap the paragraph in a div with a specific id set as anchor.
+    $current_paragraph['#theme_wrappers']['container'] = [
```
drishu: we should add the wrappers like here https://github.com/openeuropa/oe_theme/blob/3.x/oe_theme.theme#L896 so we don't have to do all this

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -0,0 +1,42 @@
+inpage_navigation_default:
+  render:
+    '#type': pattern
+    '#id': inpage_navigation
+    '#fields':
+      settings:
+        scrollspy_id: "bcl-inpage-navigation"
```
drishu: remove this

drishu (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
abel-santos-corral: I can't until v 0.16.0 is applied. ;-)

resources/js/libraries/extend/inpage_navigation.js
abel-santos-corral: We will have to get all the inpage navigation elements (all of them will have a class 'bcl-inpage-navigation') --> from there get the id of that element --> assing a new scrollspy to each id target. We'll see and have fun with that at v.0.16.0. For sure.

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -329,6 +330,51 @@ function oe_bootstrap_theme_preprocess_paragraph__oe_timeline(array &$variables)
...
+    return;
+  }
+
+  $variables['title'] = t('Page contents');
+  if (!$paragraph->get('field_oe_title')->isEmpty()) {
+    $variables['title'] = $paragraph->get('field_oe_title')->value;
+  }
+
+  if (isset($variables['content']['field_oe_title'])) {
+    unset($variables['content']['field_oe_title']);
```
drishu: remove this, rather do content.field_oe_paragraphs in the paragraph template

templates/paragraphs/paragraph--oe-content-row--variant-inpage-navigation.html.twig
```diff
@@ -0,0 +1,13 @@
...
+ * @file
+ * Theme override to display the 'Content row' #type and 'Inpage navigation' #variant paragraph.
+ *
+ * @see ./modules/contrib/paragraphs/templates/paragraph.html.twig
+ */
+#}
+{{ pattern('inpage_navigation', {
+  'title': title,
+  'links': links,
+  'content': content,
```
drishu:
```suggestion
  'content': content.field_oe_paragraphs,
```

oe_bootstrap_theme.libraries.yml
```diff
@@ -18,6 +18,10 @@ drupal.progress:
   js:
     assets/js/progress.js: {}
 
+drupal.inpage_navigation:
```
drishu: the convention is that you use drupal.something when you are overwriting a drupal core library (Ex drupal.progress), here it is not the case
```suggestion
inpage_navigation:
```

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -0,0 +1,14 @@
+{#
+/**
+ * @file
+ * Inpage navigation pattern.
+ */
+#}
+{{ attach_library('oe_bootstrap_theme/drupal.inpage_navigation') }}
```
drishu:
```suggestion
{{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
```

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
```diff
@@ -0,0 +1,197 @@
...
+        'type' => 'oe_fact',
+        'field_oe_icon' => 'box-arrow-up',
+        'field_oe_title' => '107 Tests',
+        'field_oe_subtitle' => 'Test tickets',
+        'field_oe_plain_text_long' => 'Cras vestibulum efficitur mi, quis porta tellus rutrum ut. Quisque at pulvinar sem.',
+      ]),
+    ];
+
+    $paragraph_accordion = [
+      0 => Paragraph::create([
```
drishu: numerotation like this is redundant because the keys you add are the same as default from php 0, 1, 2 ..

tests/src/Kernel/Paragraphs/ContentRowInpageNavigationTest.php
```diff
@@ -0,0 +1,197 @@
...
+      'field_oe_title' => 'Page content',
+      'field_oe_paragraphs' => $sub_paragraphs,
+      'oe_paragraphs_variant' => 'inpage_navigation',
+    ]);
+    $paragraph->save();
+
+    $html = $this->renderParagraph($paragraph);
+    $crawler = new Crawler($html);
+
+    $this->assertCount(1, $crawler->filter('div.row'));
```
drishu: I like this part where you assert the total count of elements, below this we should make a better, more clear series of assertions, so
we have 2 columns col-md-3 and col-md-9 (lets call them $links and $content)
- for links we assert that paragraphs with title have their title presnt in the $links, and we assert the count so we know there are no others that should not be;
- for the content we assert the all paragraphs are present and that the ones that have the title in $links also are wrapped here in $content 
that's it

drishu (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -0,0 +1,40 @@
...
...n pariatur ex labore qui laborum sint consequat. Aute cupidatat id sint incididunt Lorem Lorem aute aliqua sint aliquip commodo laborum quis. Ex irure sit reprehenderit in amet. Fugiat in amet proident consequat occaecat officia et consequat ipsum aute commodo. Pariatur duis consequat ipsum enim quis Lorem sint eiusmod esse. Occaecat amet esse aliquip non incididunt nisi sint do. Magna deserunt irure est nostrud deserunt in elit in enim do occaecat consequat. Incididunt duis reprehenderit commodo mollit culpa ex. Cillum ullamco cupidatat sit fugiat exercitation voluptate elit nulla duis. Dolor deserunt ad aute sint. Sunt cillum aute anim quis anim anim. Voluptate enim proident voluptate deserunt mollit. Veniam exercitation ex aliquip laborum commodo esse laboris. Veniam exercitation dolore quis ea minim magna. Labore adipisicing magna in excepteur esse ullamco aliqua tempor veniam id anim. Nulla deserunt laboris aute reprehenderit nisi quis commodo. Excepteur tempor nulla amet fugiat ullamco sit. Nostrud proident exercitation veniam elit ullamco ullamco officia minim officia commodo id duis. Laboris ullamco aliquip voluptate dolore aliquip nostrud magna ut cillum nisi officia.</p>'
```
drishu: make this smaller

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -0,0 +1,40 @@
...
+      'nav#bcl-inpage-navigation': 1
+      'h5': 1
+      'ul.nav.nav-pills.flex-column': 1
+      'li.nav-item': 4
+      'a.nav-link': 4
+      'div.col-md-9': 1
+      'h2': 4
+      'p': 8
+    equals:
+      'h2#item-1': 'Heading 1'
```
drishu: prefix with parent class so its clear where this is checked div.col-md-3

drishu (changes requested):

drishu (dismissed):

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,27 @@
+inpage_navigation:
+  label: "Inpage navigation"
```
escuriola: Use single quotes for this and below

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,27 @@
...
...ue porta id lorem sit amet efficitur. Sed id ipsum quis metus venenatis facilisis sit amet at mi.</p><h2 id='item-2'>Heading 2 with a long title going on several lines</h2><p>Ut placerat tellus dui, ut tempor orci pulvinar eget. Maecenas hendrerit risus vel tempus tincidunt. Suspendisse vitae ante nibh. Ut faucibus porttitor dui, sit amet pharetra nulla aliquam non. Donec massa mauris, convallis quis nunc nec, ultrices dapibus urna. In vel elementum ipsum, luctus volutpat odio. Maecenas aliquet justo in purus placerat dictum.</p><h2 id='item-3'>Heading 3</h2><p>Curabitur vehicula accumsan placerat. Aliquam eget tincidunt ex, sed elementum justo. Sed faucibus ultrices finibus. Vivamus pellentesque nisl ac interdum ultrices. Quisque ornare sit amet est sed pretium. Aliquam eu leo vitae nisi egestas volutpat ut non lectus. Sed condimentum tempus semper.</p><h2 id='item-4'>Heading 4</h2><p>Nullam dignissim quam tellus, eget pretium tortor vulputate quis. In vehicula enim vitae lorem fermentum convallis. Phasellus bibendum, turpis sit amet faucibus interdum, dui lorem iaculis enim, ac eleifend odio tellus eget dolor. Maecenas nunc ante, tincidunt nec risus ut, eleifend tempor nisi.</p>"
```
escuriola: Single quotes please.

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -0,0 +1,44 @@
+inpage_navigation_default:
+  render:
+    '#type': pattern
+    '#id': inpage_navigation
+    '#fields':
+      title: "Page content"
```
escuriola: Single quotes

escuriola (changes requested):

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ * Inpage navigation pattern.
+ */
+#}
+{{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
+
+{% include '@oe-bcl/bcl-inpage-navigation/inpage-navigation.html.twig' with {
+  'title': title,
+  'links': links,
+  'content': content,
+  'id': 'bcl-inpage-navigation',
```
escuriola: Is not needed because component hardcoded the id

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
```diff
@@ -0,0 +1,27 @@
+inpage_navigation:
+  label: 'Inpage navigation'
+  description: 'Inpage navigation adds a navigation element and content element so when scrolling the active element is indicated in the nav part.'
+  fields:
+    title:
+      type: 'text'
```
escuriola: In 799 I'm leaving types without quotes because of reading on IDE, the type has a different color and improve my reading. This is a convention from myself and can be discussed if you want.

escuriola (changes requested):

templates/patterns/inpage_navigation/inpage_navigation.ui_patterns.yml
escuriola: Discussed with Francesco, he is agree doing in this way.

escuriola (approved):

drishu (approved):

PR #143, opened 2021-12-16
OEL-856: [BE] Check unused paragraph fields.

abel-santos-corral:
Jira issue(s):
- [OEL-856](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-856)


abel-santos-corral:
The PR gets closed because the solution was not good

abel-santos-corral:
This became obsolete

PR #144, opened 2021-12-21
OEL-786: Prepare spacing between paragraphs.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-786

### DIFF
--- a/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
+++ b/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
@@ -27,8 +27,6 @@ use Drupal\Component\Utility\Html;
 
 /**
  * Implements hook_theme_suggestions_HOOK_alter().
- *
- * Adds markup-free template suggestion to all paragraph fields.
  */
 function oe_bootstrap_theme_paragraphs_theme_suggestions_field_alter(array &$suggestions, array $variables): void {
   $element = $variables['element'];
--- a/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
+++ b/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
@@ -26,6 +26,28 @@ use Drupal\paragraphs\Entity\Paragraph;
 use Drupal\Component\Utility\Html;
 
 /**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Adds markup-free template suggestion to all paragraph fields.
+ */
+function oe_bootstrap_theme_paragraphs_theme_suggestions_field_alter(array &$suggestions, array $variables): void {
+  $element = $variables['element'];
+
+  // Add suggestions for all 'Entity reference revisions' for each target type.
+  if ($element['#field_type'] === 'entity_reference_revisions') {
+    /** @var \Drupal\field\Entity\FieldConfig $field_definition */
+    $field_definition = $element['#items']->getFieldDefinition();
+    // Add suggestions based on the referenced entity type.
+    $suggestions[] = 'field__entity_reference_revisions__' . $field_definition->getSetting('target_type');
+    $suggestions[] = 'field__entity_reference_revisions__' . $field_definition->getSetting('target_type') . '__' . $field_definition->getName();
+
+    // Add suggestions based on the parent entity and bundle types.
+    $suggestions[] = 'field__entity_reference_revisions__' . $element['#entity_type'] . '__' . $element['#bundle'];
+    $suggestions[] = 'field__entity_reference_revisions__' . $element['#entity_type'] . '__' . $element['#bundle'] . '__' . $field_definition->getName();
+  }
+}
+
+/**
  * Implements hook_preprocess_paragraph().
  */
 function oe_bootstrap_theme_paragraphs_preprocess_paragraph__oe_links_block(array &$variables): void {
--- a/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
+++ b/modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
@@ -26,6 +26,23 @@ use Drupal\paragraphs\Entity\Paragraph;
 use Drupal\Component\Utility\Html;
 
 /**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Adds a bare, markup-free template suggestion to all paragraph fields.
+ */
+function oe_bootstrap_theme_paragraphs_theme_suggestions_field_alter(array &$suggestions, array $variables): void {
+  $element = $variables['element'];
+
+  // Do not output field labels and wrapping markup for paragraph fields.
+  if (isset($element['#entity_type']) && $element['#entity_type'] === 'paragraph') {
+    // Prepend the new suggestion to the list. This will put it right after the
+    // default field template. By doing this we allow to override single
+    // fields, while keeping all the rest markup-free.
+    array_unshift($suggestions, 'field__bare');
+  }
+}
+
+/**
  * Implements hook_preprocess_paragraph().
  */
 function oe_bootstrap_theme_paragraphs_preprocess_paragraph__oe_links_block(array &$variables): void {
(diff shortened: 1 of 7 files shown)


drishu (changes requested):
Nice, but we need to test this somehow, we have a paragraphs functional test with everything set up, we could profit and just add a test there that asserts the count of this class in the page

tests/src/Functional/ParagraphsTest.php
```diff
@@ -48,6 +48,12 @@ class ParagraphsTest extends BrowserTestBase {
   protected function setUp(): void {
     parent::setUp();
 
+    \Drupal::service('theme_installer')->install(['oe_bootstrap_theme']);
```
drishu: remove this, rather add oe_bootstrap_theme to $defaultTheme

tests/src/Functional/ParagraphsTest.php
```diff
@@ -48,6 +48,12 @@ class ParagraphsTest extends BrowserTestBase {
...
+    \Drupal::configFactory()
```
drishu: remove this

tests/src/Functional/ParagraphsTest.php
```diff
@@ -151,6 +157,49 @@ public function testSocialMediaFollowParagraph(): void {
     $this->assertSession()->pageTextContains('More channels');
   }
 
+  /**
+   * Test paragraphs are wrapped with a spacing.
+   */
+  public function testSpacingBetweenParagraphs(): void {
```
drishu: rather then doing this we could just add the div count to testLinksBlockParagraph and testSocialMediaFollowParagraph

tests/src/Functional/ParagraphsTest.php
```diff
@@ -48,6 +48,12 @@ class ParagraphsTest extends BrowserTestBase {
...
 
+    \Drupal::service('theme_installer')->install(['oe_bootstrap_theme']);
+    \Drupal::configFactory()
+      ->getEditable('system.theme')
+      ->set('default', 'oe_bootstrap_theme')
+      ->save();
+
     // Rebuild the ui_pattern definitions to collect the ones provided by
     // oe_bootstrap_theme itself.
     \Drupal::service('plugin.manager.ui_patterns')->clearCachedDefinitions();
```
drishu: try also without this line, dont think its needed

includes/menu.inc
```diff
@@ -76,16 +76,9 @@ function oe_bootstrap_theme_preprocess_links__dropbutton(&$variables) {
...
   $variables['split'] = FALSE;
-  if (isset($link['link']) && ($url = $link['link']['#url'])) {
-    $button = $link['link'];
-
-    if ($variables['split'] = $url->getRouteName() !== '<nolink>') {
-      $button['#options']['attributes']['class'][] = 'btn';
-      $button['#options']['attributes']['class'][] = 'btn-sm';
-      $button['#options']['attributes']['class'][] = 'btn-outline-primary';
-    }
+  if (isset($link['text'])) {
```
drishu: what are these changes dropbutton and dowpdown changes for ? seems out of scope of this ticket (if these are needed do report in a relevant ticket

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -24,6 +24,28 @@ use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
 use Drupal\oe_media_iframe\Plugin\media\Source\Iframe;
 use Drupal\paragraphs\Entity\Paragraph;
 
+/**
+ * Implements hook_theme_suggestions_HOOK_alter().
+ *
+ * Adds markup-free template suggestion to all paragraph fields.
```
drishu: I think this comment here makes no sense

drishu (changes requested):

includes/menu.inc
escuriola: This is needed for testing. With oe_bootstrap_theme enabled with this code when you are in a create node page  the list of available paragraphs is empty.

escuriola: Created a new ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-961

drishu (dismissed):

drishu (dismissed):

templates/overrides/field/field--entity-reference-revisions--paragraph.html.twig
```diff
@@ -0,0 +1,13 @@
+{#
+/**
+ * @file
+ * Theme override for the entity reference revisions field wich allowed to use paragraphs.
```
abel-santos-corral:
```suggestion
 * Theme override for the entity reference revisions field which allowed to use paragraphs.
```

modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.module
```diff
@@ -24,6 +24,23 @@ use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
...
+ */
+function oe_bootstrap_theme_paragraphs_theme_suggestions_field_alter(array &$suggestions, array $variables): void {
+  $element = $variables['element'];
+
+  // Do not output field labels and wrapping markup for paragraph fields.
+  if (isset($element['#entity_type']) && $element['#entity_type'] === 'paragraph') {
+    // Prepend the new suggestion to the list. This will put it right after the
+    // default field template. By doing this we allow to override single
+    // fields, while keeping all the rest markup-free.
+    array_unshift($suggestions, 'field__bare');
```
abel-santos-corral, line 41: @drishu we brought this from OE Theme, we understand is necessary in OE Bootstrap theme. Right?

Another question is that it could be good to add an assertion in the test at EntityReferenceRevisionTest. We must ensure this suggestion is applied and no labels or attributes are added (because twig field__bare is used.). Is correct this assumption?

drishu: yes, field__bare is good.
"ensure this suggestion is applied" since we check there for my-4 and find it, it means that the suggestion is applied, and that will be enough for now

tests/src/Kernel/Paragraphs/EntityReferenceRevisionTest.php
```diff
@@ -0,0 +1,113 @@
...
+
+    $build = $node->get('test_paragraphs')->view();
+    $html = $this->render($build);
+    $crawler = new Crawler($html);
+
+    $wrappers = $crawler->filter('div.my-4');
+    $this->assertCount(2, $wrappers);
+    // Assertions for field bare twig.
+    $rich_text = $wrappers->eq(1);
+    $this->assertCount(1, $rich_text);
```
drishu: if we assert above for count 2 its natural that there is a eq(1), this line is redundant

tests/src/Kernel/Paragraphs/EntityReferenceRevisionTest.php
```diff
@@ -0,0 +1,113 @@
...
+    ]);
+    $node->save();
+
+    $build = $node->get('test_paragraphs')->view();
+    $html = $this->render($build);
+    $crawler = new Crawler($html);
+
+    $wrappers = $crawler->filter('div.my-4');
+    $this->assertCount(2, $wrappers);
+    // Assertions for field bare twig.
```
drishu: checking that there is not title is fine, but this test is not about field bare, the implementation behind is not important (describing it is actually a mistake), its the resulted rendering that we check for, so please remove this comment.

tests/src/Kernel/Paragraphs/EntityReferenceRevisionTest.php
```diff
@@ -0,0 +1,113 @@
...
+    $rich_text = $wrappers->eq(1);
+    $this->assertCount(1, $rich_text);
+    $title_rich_text = $rich_text->filter('div.oe-rich-text__field-oe-title');
+    $this->assertCount(0, $title_rich_text);
+    $text_rich_text = $rich_text->filter('div.oe-rich-text__field-oe-text-long');
+    $this->assertCount(0, $text_rich_text);
```
drishu: since we don't sent the value of this field its natural that it doesn't appear no sense in checking that

tests/src/Kernel/Paragraphs/EntityReferenceRevisionTest.php
```diff
@@ -0,0 +1,113 @@
...
+    // Assertions for field entity reference revisions twig.
```
drishu: again, don't mention implementation, the behind the scenes can change, its the result that we need to check
```suggestion
    // Assert the paragraphs where rendered.
```

drishu (changes requested):

drishu (approved):

abel-santos-corral (approved):

PR #151, opened 2022-01-10
OEL-779: Review patterns: Date block, Dropdown, Grid row, Icon, Inpage navigation and Jumbotron.

abel-santos-corral:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-779

Fix patterns:
* Date block
* Dropdown
* Grid row
* Icon
* Inpage navigation
* Jumbotron


templates/patterns/dropdown/dropdown.ui_patterns.yml
```diff
@@ -1,71 +1,70 @@
...
+          path: '#'
+        - label: 'I am a button'
           button: true
         - divider: true
-        - label: "I'm a disabled button"
+        - label: 'I am a disabled button'
           button: true
           disabled: true
-        - label: "I'm a active button"
+        - label: 'I am a active button'
```
escuriola:
```suggestion
        - label: 'I am an active button'
```

templates/patterns/dropdown/dropdown.ui_patterns.yml
```diff
@@ -1,71 +1,70 @@
...
       type: select
-      label: Direction
+      label: 'Direction'
+      description: 'Sets the direction of the dropdown (if not informed, dropdown by default):'
       options:
-        dropdown: "dropdown"
-        dropup: "dropup"
-        dropend: "dropend"
-        dropstart: "dropstart"
+        dropup: 'dropup'
```
escuriola: Missing dropdown option??

templates/patterns/dropdown/dropdown.ui_patterns.yml
```diff
@@ -1,71 +1,70 @@
...
-      label: link
-      description: Show as a link
+      label: 'link'
+      description: 'When enabled, shows dropdown as a link. False by default.'
       preview: false
     direction:
       type: select
-      label: Direction
+      label: 'Direction'
+      description: 'Sets the direction of the dropdown (if not informed, dropdown by default):'
```
escuriola: Add the default value in the options intead here

templates/patterns/dropdown/pattern-dropdown.html.twig
```diff
@@ -14,4 +14,4 @@
   'remove_wrapper': remove_wrapper,
```
escuriola: We have a problem with trigger. He is waiting a button or link "component" but nobody has to know how is a component internally so we need the user here put a button or link pattern and map to adapt to component.

templates/patterns/icon/icon.ui_patterns.yml
```diff
@@ -1,41 +1,40 @@
...
+        envelope: 'envelope'
+        eye: 'eye'
+        hand-index: 'info'
+      preview: 'eye'
     size:
-      type: "select"
-      label: "Size of icon"
+      type: select
+      label: 'Size of icon'
+      description: 'A list of icon sizes ("s" by default):'
```
escuriola: Add the default in the option list instead here

escuriola (changes requested):

templates/patterns/dropdown/dropdown.ui_patterns.yml
abel-santos-corral: I removed because I checked how the pattern defines the field. But adding it back, let's see if it makes more sense (and added the default comment in the option)

templates/patterns/dropdown/dropdown.ui_patterns.yml
```diff
@@ -1,71 +1,66 @@
...
+        dropdown-menu-end: 'dropdown-menu-end'
+        dropdown-menu-sm-end: 'dropdown-menu-sm-end'
+        dropdown-menu-md-end: 'dropdown-menu-md-end'
+        dropdown-menu-lg-end: 'dropdown-menu-lg-end'
+        dropdown-menu-xl-end: 'dropdown-menu-xl-end'
+        dropdown-menu-xxl-end: 'dropdown-menu-xxl-end'
+        dropdown-menu-start: 'dropdown-menu-start'
+        dropdown-menu-sm-start: 'dropdown-menu-sm-start'
+        dropdown-menu-md-start: 'dropdown-menu-md-start'
+        dropdown-menu-lg-start: 'dropdown-menu-lg-start'
```
brummbar: Too complex, we don't need all of this. IF, and it's a big if, is ever needed, these actually map to classes and they can be passed together with attributes.

templates/patterns/icon/icon.ui_patterns.yml
```diff
@@ -1,41 +1,40 @@
 icon:
-  label: "Icon"
-  description: "Icons available here: https://icons.getbootstrap.com/"
+  label: 'Icon'
+  description: 'A component to show an icon. For more information visit: https://icons.getbootstrap.com/.'
   settings:
```
brummbar: This is all wrong. Settings are addictional to fields. We have no fields in this pattern, it's not ok. These are all fields of an icon, except transformation which is a setting.

brummbar (changes requested):

brummbar (approved):

escuriola (approved):

PR #153, opened 2022-01-12
OEL-779: Review patterns: Link, Links block, List,List group and Modal

abel-santos-corral:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-779

Fix patterns:
* Link
* Links block
* List
* List group
* Modal


includes/pattern.inc
```diff
@@ -139,6 +140,20 @@ function oe_bootstrap_theme_preprocess_pattern_carousel(array &$variables) {
   // @endcode
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_link(array &$variables): void {
+  $variables['link_id'] = Html::getUniqueId('bcl-link');
```
brummbar: Nono we shouldn't do this. Like this we are forcing an id to be present in ALL the links we render, which is bad. Plus the bcl template can use the attributes one for this, so no need.

templates/patterns/list_group/list_group.ui_patterns.yml
```diff
@@ -1,53 +1,52 @@
...
+        unordered: 'unordered (default)'
+        ordered: 'ordered'
+        actionable: 'actionable'
     flush:
       type: boolean
-      label: Flush
-      description: Add flush to remove some borders and rounded corners to render list group items edge-to-edge in a parent container (e.g., cards).
+      label: 'Flush'
+      description: 'When enabled, adds flush to remove some borders and rounded corners to render list group items edge-to-edge in a parent container (e.g., cards). False by default.'
     horizontal:
```
brummbar: Let's remove this too.

templates/patterns/list_group/pattern-list-group.html.twig
```diff
@@ -1,12 +1,24 @@
...
+{% set _items = [] %}
+
+{% for item in items %}
+  {% if item.html %}
+    {% set item = item|merge({
+      'variant': item.link_style
+    }) %}
+  {% endif %}
+  {% set _items = _items|merge([item]) %}
+{% endfor %}
```
brummbar: I think this is wrong, we are not setting an API here. Any change in the component will change the pattern, which is not good.

templates/patterns/modal/pattern-modal.html.twig
```diff
@@ -1,31 +1,28 @@
...
 /**
  * @file
- * Modal component.
+ * Modal pattern.
  */
 #}
-
-{% set _modal_id = modal_id ? 'modal_' ~ modal_id : 'modal' %}
-
 {% if btn_label is not empty %}
```
brummbar: I don't remember this, so we are having a way to render button and modal together?

templates/patterns/modal/modal.ui_patterns.yml
```diff
@@ -1,78 +1,75 @@
...
+      label: 'Button label'
+      description: 'The button text to trigger the modal. If is not present, then button will not be printed, only the modal container.'
+      preview: 'Open the modal'
     header:
       type: text
-      label: "Header"
-      description: "Modal header"
+      label: 'Header'
+      description: 'The modal’s header.'
       preview:
```
brummbar: As I wrote for some other pattern, we should give an API. Instead of header, we should have "title" and wrap the title text in it. Same for button. @escuriola maybe remembers which other pattern had the same remark.

brummbar (changes requested):

includes/pattern.inc
```diff
@@ -181,11 +181,18 @@ function oe_bootstrap_theme_preprocess_pattern_modal(array &$variables): void {
...
+  // Convert to array when only the icon name has been passed.
+  if (!is_array($variables['icon'])) {
     $variables['icon'] = [
       'name' => $variables['icon'],
-      'path' => $variables['bcl_icon_path'],
     ];
   }
 
+  $variables['icon'] += [
+    'path' => $variables['bcl_icon_path'],
```
escuriola, line 196: Question: If icon pattern has already hardcoded the path with the bcl_icon_path, do we really need this?

templates/patterns/link/link.ui_patterns.yml
```diff
@@ -34,7 +34,7 @@ link:
       description: 'The button URL (optional).'
       preview: 'http://example.com'
     icon:
-      type: text
-      label: 'Name of the icon'
-      description: 'The name of the icon (optional).'
+      type: mixed
+      label: 'Icon'
+      description: 'An icon name or an icon array definition (optional).'
```
escuriola, line 39: All other patterns we change to icon name only. Why here allow array and name?

tests/src/Kernel/fixtures/markup_rendering_patterns/link.yml
```diff
@@ -79,7 +81,8 @@ link_pattern_disabled_with_icon_after_without_removing_icon_spaces:
         icon_position: 'after'
       label: 'Link to example'
       path: 'https://example.com'
-      icon: 'alarm'
+      icon:
```
escuriola, line 84: Test icon as icon name also

escuriola (approved):

tests/src/Kernel/fixtures/markup_rendering_patterns/link.yml
```diff
@@ -80,49 +81,49 @@ link_pattern_secondary_disabled_with_icon_after_without_removing_icon_spaces:
       path: 'https://example.com'
       icon: 'alarm'
   assert:
-    contains:
-      'svg': '#alarm'
-      'a[data-drupal-selector=link]': 'Link to example'
+    ccontains:
```
abel-santos-corral: Is contains. Check this

brummbar (approved):

Issue #156, opened 2022-01-14
Preprocess of menus is breaking menus with contrib templates


dxvargas:
### Description

When installing a module having a menu and a template already and using that menu, an error may occur:
`InvalidArgumentException: The URI '' is invalid. You must use a valid URI scheme. in Drupal\Core\Url::fromUri() (line 291 of core/lib/Drupal/Core/Url.php). `

This is because regular menu templates have items with a title and a link.
But currently the template will not get items with these entries, since after #74 these are replaced with the entries label and path.
Since there is no link entry, `fromUri()` will break after.
Even without this error, the menu entries won't get displayed correctly, since the template is not getting the expected items title and link.

### Steps to reproduce:

This error can be seen following these steps:
1. Require the contrib module `menu_item_extras`;
2. Enable it together with "Menu Item Extras Demo Base";
3. Place the "Menu Item Extras Demo Menu" block in the section "Navbar branding";
4. Check the front page, the error will show.

This is because the mentioned menu has already their own templates. When trying to use them, the described problem happens.

dxvargas:
I propose a simple solution:

The code for adapting the menu to the templates of `oe_bootstrap_theme` are better placed in a preprocess for the nav pattern, instead of a preprocess of the menus. Like this, only the menus using the nav pattern will be affected by the changes needed in this theme.

dxvargas:
Another possibility is to adapt the preprocess of the menus to not remove the existing items fields (title, url, attributes), but to only add the needed ones for the BCL templates (label, path, attributes).

brummbar:
Marking as solved.

PR #167, opened 2022-01-27
OEL-1057: Fix issue with menu preprocess overriding core keys.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1057

### DIFF
--- a/includes/menu.inc
+++ b/includes/menu.inc
@@ -8,6 +8,7 @@
 declare(strict_types = 1);
 
 use Drupal\Core\Template\Attribute;
+use Drupal\Core\Url;
 
 /**
  * Implements hook_preprocess_menu().
@@ -17,41 +18,35 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
     return;
   }
 
-  $items = $variables['items'];
-  foreach ($items as $key => $item) {
-    $class = $item['in_active_trail'] ? ['active'] : [];
+  $variables['pattern_items'] = [];
+  foreach ($variables['items'] as $key => $item) {
     $menu_item = [
       'label' => $item['title'],
-      'path' => $item['url']->toString(),
-      'attributes' => new Attribute(['class' => $class]),
+      'path' => $item['url'] instanceof Url ? $item['url']->toString() : '',
+      'attributes' => new Attribute(['class' => $item['in_active_trail'] ? ['active'] : []]),
     ];
-    $variables['items'][$key] = $menu_item;
-
     if (empty($item['below'])) {
+      $variables['pattern_items'][$key] = $menu_item;
       continue;
     }
 
-    $class[] = 'nav-link';
-    $menu_item['attributes'] = new Attribute(['class' => $class]);
-    $item_with_submenu = [
+    $submenu = $menu_item;
+    $submenu['attributes'] = new Attribute(['class' => 'nav-link']);
+    $menu_item += [
       'standalone' => TRUE,
       'dropdown' => TRUE,
       'link' => TRUE,
-      'alignment' => 'dropdown-menu-sm-start',
-      'trigger' => $menu_item,
+      'alignment' => 'dropdown-menu_item-sm-start',
+      'trigger' => $submenu,
     ];
-    $variables['items'][$key] = $item_with_submenu;
-
     foreach ($item['below'] as $key_sub_item => $sub_item) {
-      $subclass = $sub_item['in_active_trail'] ? ['active'] : [];
-      $sub_item_menu = [
-        'label' => $sub_item['title'],
-        'path' => $sub_item['url']->toString(),
-        'attributes' => new Attribute(['class' => $subclass]),
-      ];
-      $variables['items'][$key]['items'][$key_sub_item] = $sub_item_menu;
+      $menu_item['items'][$key_sub_item]['label'] = $sub_item['title'];
+      $menu_item['items'][$key_sub_item]['path'] = $sub_item['url'] instanceof Url ? $sub_item['url']->toString() : '';
+      $menu_item['items'][$key_sub_item]['attributes'] = new Attribute(['class' => $sub_item['in_active_trail'] ? ['active'] : []]);
     }
+    $variables['pattern_items'][$key] = $menu_item;
   }
+
 }
 
 /**
--- a/includes/menu.inc
+++ b/includes/menu.inc
@@ -20,7 +20,7 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
 
   $variables['pattern_items'] = [];
   foreach ($variables['items'] as $key => $item) {
-    $menu_item = [
+    $pattern_item = [
       'label' => $item['title'],
       'path' => $item['url'] instanceof Url ? $item['url']->toString() : '',
       'attributes' => new Attribute([
@@ -28,13 +28,13 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
       ]),
     ];
     if (empty($item['below'])) {
-      $variables['pattern_items'][$key] = $menu_item;
+      $variables['pattern_items'][$key] = $pattern_item;
       continue;
     }
 
-    $submenu = $menu_item;
+    $submenu = $pattern_item;
     $submenu['attributes'] = new Attribute(['class' => 'nav-link']);
-    $menu_item += [
+    $pattern_item += [
       'standalone' => TRUE,
       'dropdown' => TRUE,
       'link' => TRUE,
@@ -42,7 +42,7 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
       'trigger' => $submenu,
     ];
     foreach ($item['below'] as $key_sub_item => $sub_item) {
-      $menu_item['items'][$key_sub_item] = [
+      $pattern_item['items'][$key_sub_item] = [
         'label' => $sub_item['title'],
         'path' => $sub_item['url'] instanceof Url ? $sub_item['url']->toString() : '',
         'attributes' => new Attribute([
@@ -50,7 +50,7 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
         ]),
       ];
     }
-    $variables['pattern_items'][$key] = $menu_item;
+    $variables['pattern_items'][$key] = $pattern_item;
   }
 
 }
--- a/includes/menu.inc
+++ b/includes/menu.inc
@@ -42,9 +42,13 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
       'trigger' => $submenu,
     ];
     foreach ($item['below'] as $key_sub_item => $sub_item) {
-      $menu_item['items'][$key_sub_item]['label'] = $sub_item['title'];
-      $menu_item['items'][$key_sub_item]['path'] = $sub_item['url'] instanceof Url ? $sub_item['url']->toString() : '';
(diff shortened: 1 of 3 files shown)


includes/menu.inc
```diff
@@ -18,11 +18,6 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
   $items = $variables['items'];
   foreach ($items as $key => $item) {
```
abel-santos-corral, line 19: Please, Carlos, check if this can be done:
foreach ($variables['items'] as $key => $item) {

abel-santos-corral (commented):
Plese, check if we can only have one commit instead of 2.

abel-santos-corral (dismissed):

includes/menu.inc
```diff
@@ -18,36 +18,33 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
       'standalone' => TRUE,
       'dropdown' => TRUE,
       'link' => TRUE,
       'alignment' => 'dropdown-menu-sm-start',
-      'trigger' => $menu_item,
+      'trigger' => [
+        'label' => $item['title'],
+        'path' => $item['url']->toString(),
+        'attributes' => new Attribute(['class' => $class]),
+      ],
```
donquixote: @escuriola I see some repetition here vs above.

And I wonder, where is this 'trigger' value used, and which array keys does it expect?
In the `menu.html.twig`, I see we are directly calling the pattern.
In `pattern-navigation.html.twig` we are sending the items directly to `bcl-navigation.html.twig`.

Here they go to `bcl-dropdown.html.twig` which indeed uses the 'trigger' part.
The values within 'trigger' are then sent to `bcl-link.html.twig` or `bcl-button.html.twig`. At least 'link' expects 'path' and 'label', whereas 'button' only expects 'label'.

includes/menu.inc
```diff
@@ -18,36 +18,33 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
       $subclass = $sub_item['in_active_trail'] ? ['active'] : [];
-      $sub_item_menu = [
-        'label' => $sub_item['title'],
-        'path' => $sub_item['url']->toString(),
-        'attributes' => new Attribute(['class' => $subclass]),
-      ];
-      $variables['items'][$key]['items'][$key_sub_item] = $sub_item_menu;
+      $variables['items'][$key]['items'][$key_sub_item]['label'] = $sub_item['title'];
+      $variables['items'][$key]['items'][$key_sub_item]['path'] = $sub_item['url']->toString();
+      $variables['items'][$key]['items'][$key_sub_item]['attributes'] = new Attribute(['class' => $subclass]);
```
donquixote: @escuriola
This change seems to be pointless.
The original Drupal menu item does not have any 'items' key that we were destroying. 'items' is a completely new key that is only understood by bcl-navigation.

includes/menu.inc
```diff
@@ -18,36 +18,33 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
       'alignment' => 'dropdown-menu-sm-start',
-      'trigger' => $menu_item,
+      'trigger' => [
+        'label' => $item['title'],
+        'path' => $item['url']->toString(),
+        'attributes' => new Attribute(['class' => $class]),
+      ],
     ];
-    $variables['items'][$key] = $item_with_submenu;
+    $variables['items'][$key] += $item_with_submenu;
```
donquixote: @escuriola 
Overall I feel we have a lot of redundant juggling of variables here:
We are having local variables like `$item` from the foreach, `$items` from outside the foreach, `$item_with_submenu`, and in the old code we had `$menu_item`. But then we are writing each value to `$variables['items'][$key]`.

We have two options that would simplify this:
- By-reference `foreach ($variables['items'] as $key => &$item) {`, then write everything directly to `$item`. Perhaps we could call it `$item_ref` to remind the reader that this is by-reference.
- Don't use by-reference, write everything to `$item`, and in the end set `$variables['items'][$key] = $item;`.

Also, instead of local vars like `$item_with_submenu` with later `+=`, we could directly call `$item += [...]`.

----

I have a more general concern with the entire rewriting of variables here, but I will address this in another comment.

donquixote:
A general concern I have with this entire rewriting mechanism:
(this also applies to oe_theme, so perhaps we need feedback from Francesco on this)

### LSP for templates and patterns

In my opinion, we should treat patterns and Drupal theme template overrides in the same way we treat method overrides in OOP code: We should respect the parameter signature of the parent implementation. Liskov Substitution Principle (just adding this to easily find this post later).

The variables sent to the template should have roughly the same array structure as the variables that are sent to the respective core template. In preprocess we can do some minor changes, but we should not fundamentally change the structure of the array.

Not doing this leads to problems with other components (modules) that try to inject their own templates - as we have seen here.

With this PR we are trying to work around this problem by building a mixed structure that is understood both by our templates and by core templates.

This mixed structure has lots of duplication, e.g. now we have `$item['title']` and `$item['label']` essentially doing the same thing.
Any other preprocess hook that tries to alter the title would have to alter both 'label' and 'title' to work for all templates.

#### Theme hook signature vs template signature
In Drupal it is a little bit more complicated than that.

The signature of a theme hook is any variables we sent to the theme hook, e.g. from a render element.
The signature of a template is any variables expected or understood by the template.
These two signatures can differ, if there is a lot of rewriting going on in the default preprocess functions for that theme hook.

The goal would be to respect the spirit of LSP both for the theme hook signature _and_ for the template signature.

Perhaps the LSP-ish rules are a bit vague for templates and theme hooks, but we should at least not completely rewrite these 
arrays.

#### Patterns vs BCL

Btw the same applies to patterns vs bcl templates:

The variables received by a pattern variant should always be the same as for the main variant, and as defined in the pattern definition (yml file).

The variables sent to the BCL template can have a different structure.

The rewriting has to happen in the pattern template, possibly with the help of a twig function.

### Solution

The boundary from "Drupal menu theme hook" to "navigation pattern" is in the menu template itself, where we include the pattern.
Here we need to convert all variables from a structure understood by menu templates into a structure understood by the navigation pattern.

To not repeat this in every menu template, we could write a twig function `drupal_menu_to_navigation_pattern()`.

So in `menu.html.twig` we would have

```twig
{{ pattern('navigation', {
  items: items|drupal_menu_to_navigation_pattern
}) }}
```

This twig function could then do a very clean conversion, and would not need to support a duplicate structure.

### Next steps
Don't rewrite everything yet.
What I wrote above is my personal opinion, and I need to align with @brummbar and @drishu.

Perhaps we want to deliver a quick fix that keeps most of the logic in the preprocess function.

donquixote:
I talked with @brummbar . He rightly suggested we should use a separate variable for the items that are sent to the pattern.
E.g. `$variables['pattern_items']`.
This would solve most of the problems.

includes/pattern.inc
```diff
@@ -161,3 +163,46 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
 }
+
+/**
+ * Implements hook_preprocess_menu().
```
donquixote: That's not correct.

donquixote: I suggest "Impelements hook_preprocess_HOOK() for 'pattern_navigation'.". But look at other functions how we do it there.

includes/menu.inc
```diff
@@ -17,41 +18,35 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
-  foreach ($items as $key => $item) {
-    $class = $item['in_active_trail'] ? ['active'] : [];
+  $variables['pattern_items'] = [];
+  foreach ($variables['items'] as $key => $item) {
     $menu_item = [
       'label' => $item['title'],
-      'path' => $item['url']->toString(),
-      'attributes' => new Attribute(['class' => $class]),
+      'path' => $item['url'] instanceof Url ? $item['url']->toString() : '',
+      'attributes' => new Attribute(['class' => $item['in_active_trail'] ? ['active'] : []]),
```
donquixote: Long line.
I see that this was previously done with a local variable `$class`.
In my previous review I mentioned that we do too much "juggling with variable". But I don't think I meant this one.

Easiest fix is to simply introduce line breaks, like so:

```php
      'attributes' => new Attribute([
        'class' => $item['in_active_trail'] ? ['active'] : [],
      ]),
```
We can debate whether or not we should use a local variable here, and whether we should omit the 'attributes' altogether if we are not adding any class. But it seems good enough to me with just line breaks.

includes/menu.inc
```diff
@@ -17,41 +18,41 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
-  foreach ($items as $key => $item) {
-    $class = $item['in_active_trail'] ? ['active'] : [];
-    $menu_item = [
+  $variables['pattern_items'] = [];
+  foreach ($variables['items'] as $key => $item) {
+    $pattern_item = [
       'label' => $item['title'],
-      'path' => $item['url']->toString(),
-      'attributes' => new Attribute(['class' => $class]),
+      'path' => $item['url'] instanceof Url ? $item['url']->toString() : '',
```
donquixote: @escuriola Is `''` a good fallback value if there is no url?
When does this happen, and what is the effect of putting '' here? (instead of e.g. null) Can we reproduce this?

donquixote (dismissed):

templates/overrides/menu/menu.html.twig
```diff
@@ -4,5 +4,6 @@
  */
 #}
 {{ pattern('navigation', {
-  'items': items
+  'variant': 'pills',
```
drishu, line 7: why do we use pills ?

drishu (commented):
I agree with the approach, doing this in a pattern preprocess would have been wrong architecturally because the pattern preprocess and pattern template should expect the values as defined in the yml file.

drishu:
resolves https://github.com/openeuropa/oe_bootstrap_theme/issues/156

templates/overrides/menu/menu.html.twig
escuriola: At netlify the default navigation has pills: https://oelibrary.netlify.app/default/?path=/story/components-navigation--default&args=alignment:!undefined

drishu: Well, since there is not specific requirement we can go with that

includes/menu.inc
drishu: looking into it, it's always an Url object, even if we use `<nolink>` route, the result of toString of that is indeed an empty string, BUT since the field is core required, its always gonna be a Url object, this wont change.
So we don't need this default value
```
'path' => $item['url']->toString();
```

drishu (changes requested):

drishu (dismissed):

includes/menu.inc
donquixote: I see now where the 'trigger' key is used.
Imo this is a bit weird: By setting `$pattern_item['trigger']`, some other keys within `$pattern_item` such as `$pattern_item['label']` no longer have any effect.
Trying to clean this up would be out of scope in this issue. So let's keep it as it is.

donquixote (dismissed):

drishu (approved):

donquixote (approved):
Still ok after rebase.

PR #169, opened 2022-01-31
OEL-836: Content banner pattern

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-836

### DIFF
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -161,4 +161,3 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
 }
-
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -10,7 +10,6 @@ declare(strict_types = 1);
 use Drupal\Component\Utility\Html;
 use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
 use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObjectInterface;
-use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
  * Implements hook_preprocess_HOOK() for badge_preview patterns.
@@ -163,42 +162,3 @@ function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): v
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
 }
 
-/**
- * Implements hook_preprocess_HOOK() for content banner.
- */
-function oe_bootstrap_theme_preprocess_pattern_content_banner(array &$variables) {
-  oe_bootstrap_theme_get_image($variables);
-}
-
-/**
- * Get imageValueObject from variables.
- *
- * @param array $variables
- *   The render array.
- */
-function oe_bootstrap_theme_get_image(array &$variables): void {
-  if (isset($variables['date'])) {
-    $date = DrupalDateTime::createFromFormat('Y-m-d', $variables['date']);
-    $variables['date'] = $date->format('j M Y');
-  }
-
-  if (!isset($variables['image'])) {
-    return;
-  }
-
-  // This conversion is needed only to run markup rendering tests, as they
-  // are reading from a yaml file where we cannot specify objects.
-  // Do not ever rely on this for your patterns.
-  // @todo Remove this! Code duplication below is intentional, these lines are
-  //   to be removed, do not refactor unless to delete them.
-  if (is_array($variables['image']) && !empty($variables['image'])) {
-    $variables['image'] = ImageValueObject::fromArray($variables['image']);
-  }
-
-  if ($variables['image'] instanceof ImageValueObjectInterface) {
-    // Transform an object implementing ImageValueObjectInterface to our own
-    // implementation, so that we can use the specific method.
-    $variables['image'] = ImageValueObject::fromArray($variables['image']->getArray())
-      ->toRenderArray();
-  }
-}
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -10,6 +10,7 @@ declare(strict_types = 1);
 use Drupal\Component\Utility\Html;
 use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
 use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObjectInterface;
+use Drupal\Core\Datetime\DrupalDateTime;
 
 /**
  * Implements hook_preprocess_HOOK() for badge_preview patterns.
@@ -160,4 +161,44 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
  */
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
+}
+
+/**
+ * Implements hook_preprocess_HOOK() for content banner.
+ */
+function oe_bootstrap_theme_preprocess_pattern_content_banner(array &$variables) {
+  oe_bootstrap_theme_get_image($variables);
+}
+
+/**
+ * Get imageValueObject from variables.
+ *
+ * @param array $variables
+ *   The render array.
+ */
+function oe_bootstrap_theme_get_image(array &$variables): void {
+  if (isset($variables['date'])) {
+    $date = DrupalDateTime::createFromFormat('Y-m-d', $variables['date']);
+    $variables['date'] = $date->format('j M Y');
+  }
+
+  if (!isset($variables['image'])) {
+    return;
+  }
+
+  // This conversion is needed only to run markup rendering tests, as they
+  // are reading from a yaml file where we cannot specify objects.
+  // Do not ever rely on this for your patterns.
+  // @todo Remove this! Code duplication below is intentional, these lines are
+  //   to be removed, do not refactor unless to delete them.
+  if (is_array($variables['image']) && !empty($variables['image'])) {
+    $variables['image'] = ImageValueObject::fromArray($variables['image']);
+  }
+
+  if ($variables['image'] instanceof ImageValueObjectInterface) {
+    // Transform an object implementing ImageValueObjectInterface to our own
+    // implementation, so that we can use the specific method.
+    $variables['image'] = ImageValueObject::fromArray($variables['image']->getArray())
+      ->toRenderArray();
+  }
 }
--- a/templates/patterns/content_banner/content-banner.ui_patterns.yml
+++ b/templates/patterns/content_banner/content-banner.ui_patterns.yml
@@ -6,13 +6,14 @@ content_banner:
         type: string
         label: 'Background'
         description: 'Set a gray or white background at the content banner. Defaults to white.'
-        preview: 'white'
+        escape: false
+        preview: 'gray'
       image:
         type: ImageValueObject
         label: 'Image'
         description: 'Image of the content.'
         preview:
-          src: 'https://placeimg.com/1200/600/tech'
+          src: 'https://placeimg.com/255/255/tech'
           alt: 'Alternative text for content banner image'
       title:
         type: text
@@ -23,12 +24,12 @@ content_banner:
         type: text
         label: 'Content'
         description: 'Text with the summary of the content.'
-        preview: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.'
-      creation_date:
+        preview: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec. Sed facilisis dictum vestibulum. Fusce commodo hendrerit diam, pretium tempus leo varius sit amet. Etiam interdum, orci at sagittis luctus, lorem libero tempus mauris, a fermentum libero orci semper lacus. Duis tristique fringilla magna, eu egestas dolor molestie non.'
+      date:
         type: DateTime
-        label: 'Creation date'
-        description: 'Creation date of the content.'
-        preview: ''
+        label: 'Date'
+        description: 'Date of the content.'
+        preview: '2021-05-26'
       badges:
         type: array
         label: 'Badges'
--- a/templates/patterns/content_banner/content-banner.ui_patterns.yml
+++ b/templates/patterns/content_banner/content-banner.ui_patterns.yml
@@ -1,0 +1,40 @@
+content_banner:
+    label: 'Content banner'
+    description: 'Shows the content summary along with the title, image and badges.'
+    fields:
+      background:
+        type: string
+        label: 'Background'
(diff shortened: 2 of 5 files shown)


templates/patterns/content_banner/content-banner.ui_patterns.yml
```diff
@@ -0,0 +1,40 @@
...
+        type: text
+        label: 'Title'
+        description: 'Title of the content.'
+        preview: 'This is the title of this page'
+      content:
+        type: text
+        label: 'Content'
+        description: 'Text with the summary of the content.'
+        preview: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.'
+      creation_date:
```
brummbar: Just date.

templates/patterns/content_banner/content-banner.ui_patterns.yml
```diff
@@ -0,0 +1,40 @@
...
+        type: DateTime
+        label: 'Creation date'
+        description: 'Creation date of the content.'
+        preview: ''
```
brummbar: Pass a date, and transform it in the related hook_somethingsomething_preview().

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -0,0 +1,39 @@
+{#
+/**
+ * @file
+ * Content banner pattern.
+ */
+#}
+{% set _content %}
+  {% if creation_date is not empty %}
+    <div class="mb-4 mt-4">
+      <span class="text-muted me-3"> {{ creation_date }} </span>
```
brummbar:
```suggestion
      <span class="text-muted me-3">{{ creation_date }}</span>
```

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -0,0 +1,39 @@
...
+    </div>
+  {% endif %}
+  <p> {{ content }} </p>
+{% endset %}
+
+{% set _image = {
+  'path': image.src,
+  'alt': image.alt,
+} %}
```
brummbar: Why this extra variable here, when we can just pass this below?

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -0,0 +1,39 @@
...
+ * Content banner pattern.
+ */
+#}
+{% set _content %}
+  {% if creation_date is not empty %}
+    <div class="mb-4 mt-4">
+      <span class="text-muted me-3"> {{ creation_date }} </span>
+    </div>
+  {% endif %}
+  <p> {{ content }} </p>
```
brummbar:
```suggestion
  <div>{{ content }}</div>
```

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,111 @@
...
+      title: 'This is the title of this page'
+      content: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.'
+      badges:
+        - label: 'Primary badge'
+          rounded_pill: true
+        - label: 'Secondary badge'
+          rounded_pill: true
+  assertions:
+    count:
+      'div.bcl-content-banner.pt-lg-4-75.pt-3': 1
```
brummbar: pt-lg-4-75.pt-3 is not important for us

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,111 @@
...
+      'div.bcl-content-banner.bg-white': 1
+      'div.bcl-content-banner > div.container > div.row > div.card.bg-transparent.border-0': 1
```
brummbar: What are we asserting with this line?

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,111 @@
...
+      'div.card > div.row > div.col-md-4.col-lg-3 > img[src="https://placeimg.com/1200/600/tech"]': 1
+      'div.card > div.row > div.col-md-8.col-lg-9.col-xxl-8 > div.card-body.pt-md-0.px-0.px-md-3.pb-0': 1
```
brummbar: What are we asserting with this line?

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,111 @@
...
+      'div.card-body > h2': 1
+      'div.card-body > p': 1
+      'div.card-body > div.mt-2-5 > span.badge': 2
```
brummbar: Most likely we can drop div.mt-2-5

templates/patterns/content_banner/content-banner.ui_patterns.yml
```diff
@@ -0,0 +1,40 @@
...
+        preview: 'This is the title of this page'
+      content:
+        type: text
+        label: 'Content'
+        description: 'Text with the summary of the content.'
+        preview: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.'
+      creation_date:
+        type: DateTime
+        label: 'Creation date'
+        description: 'Creation date of the content.'
```
brummbar: Just a date, not strictly the creation.

brummbar (changes requested):
Didn't manually test yet.

templates/patterns/content_banner/pattern-content-banner.html.twig
donquixote: We have some places where this is not a date, but something else, and it can contain multiple values with space between.
https://oelibrary.netlify.app/default/?path=/story/features-group--landing
https://oelibrary.netlify.app/default/?path=/story/features-users--view (it was decided this should use content banner instead of card for the top region)

Let's name this parameter "meta" and allow multiple values, which then get wrapped in span with horizontal margin, ok?

includes/pattern.inc
```diff
@@ -160,3 +163,43 @@ function _oe_bootstrap_theme_extract_carousel_image(&$item) {
...
+  oe_bootstrap_theme_get_image($variables);
+}
+
+/**
+ * Get imageValueObject from variables.
+ *
+ * @param array $variables
+ *   The render array.
+ */
+function oe_bootstrap_theme_get_image(array &$variables): void {
```
donquixote: This function should be renamed or explained better, imo.
The current name is quite generic, but I assume this is only meant for a very specific "$variables" array.
Also, I doubt that "$variables" is really a render array. Or is it?
Also, the name "get_image" suggests that we are returning a value (an image), when in fact we are writing into the $variables array.

donquixote: Or perhaps we don't need this function at all, and can do it all in the preprocess.

donquixote: I see two options:
1. Change this into a function that receives the value from `$variables['image']` as parameter, and returns the new value for `$variables['image']`. (or we can store it in a different key)
2. Make this a twig function that can be called from the template: `image|bcl_image`, instead of doing it from preprocess.

@brummbar opinion?

templates/patterns/content_banner/content-banner.ui_patterns.yml
```diff
@@ -0,0 +1,41 @@
...
+    description: 'Shows the content summary along with the title, image and badges.'
+    fields:
+      background:
+        type: string
+        label: 'Background'
+        description: 'Set a gray or white background at the content banner. Defaults to white.'
+        escape: false
+        preview: 'gray'
+      image:
+        type: ImageValueObject
```
donquixote: Technically, we also support an array of the format `array{src: string, alt: string}`.
Should we document this? Or is this a hidden feature we don't want people to use?

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -0,0 +1,36 @@
...
+  'title': {
+    'content': title,
+  },
+  'content': _content,
+  'image': {
+    'path': image['#uri'],
+    'alt': image['#alt'],
+  },
+  'badges': _badges,
+  'attributes': attributes,
```
donquixote, line 40: I don't think we need the quotes around the keys.

```twig
  background: background,
  title: {
    content: title,
  },
  content: _content,
  image: {
    path: image['#uri'],
    alt: image['#alt'],
  },
  badges: _badges,
  attributes: attributes,
```

includes/pattern.inc
```diff
@@ -160,3 +163,43 @@ function _oe_bootstrap_theme_extract_carousel_image(&$item) {
...
+  if (is_array($variables['image']) && !empty($variables['image'])) {
+    $variables['image'] = ImageValueObject::fromArray($variables['image']);
+  }
+
+  if ($variables['image'] instanceof ImageValueObjectInterface) {
+    // Transform an object implementing ImageValueObjectInterface to our own
+    // implementation, so that we can use the specific method.
+    $variables['image'] = ImageValueObject::fromArray($variables['image']->getArray())
+      ->toRenderArray();
+  }
```
donquixote: Why create a render array, which we have to convert again later in the template?

includes/pattern.inc
brummbar: I think the refactoring done around this function is very wrong, we are handling two different cases at once.
Some notes aside from that:
* let's preprend `_` to "internal" functions, it's an ugly Drupalism but quite effective in this realm
* try to pass a simple parameter (like array) and return an array, instead of passing variables by reference.

templates/patterns/content_banner/content-banner.ui_patterns.yml
brummbar: No when we use ImageValueObject we should support only that.

templates/patterns/content_banner/pattern-content-banner.html.twig
brummbar: True, but we added them pretty much everywhere. Just a choice, not needed indeed though.

includes/pattern.inc
brummbar: Yes this is wrong due to a bad refactoring.

includes/pattern.inc
```diff
@@ -134,30 +134,53 @@ function oe_bootstrap_theme_preprocess_pattern_navbar(array &$variables): void {
...
       ->toRenderArray();
   }
-}
 
-/**
- * Implements hook_preprocess_HOOK() for inpage navigation pattern.
- */
-function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
-  $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
+  return $image;
```
brummbar: Please revert all of the changes to other functions.
Then let's focus on what is supposed to be passed to the content banner: an ImageValueObject instance.
What is needed in the BCL template? Path and alt. So there's no need for any preprocess. We just do `image.src` and `image.alt`. See the [usual oe_theme](https://github.com/openeuropa/oe_theme/blob/0877a9568616af27518ffa3beb34514520fc60ae/templates/patterns/text_featured_media/pattern-text-featured-media.html.twig#L48).
In card we did different, because the card BCL template needs markup. So in the pattern preprocess we make the transformation from the expected pattern variable (again an object) to the final render array.

**Additional notes**:
We could, in the future, ditch even the transformation in render array done in the preprocess, as we can invoke directly the method in twig. The thing is that at this moment I added the `toRenderArray()`  method only to our implementation of ImageValueObjectInterface, so we need to do some conversion. This is ideally temporary, and we should add the method to the interface.

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,115 @@
...
+      badges:
+        - label: 'Primary badge'
+          rounded_pill: true
+        - label: 'Secondary badge'
+          rounded_pill: true
+  assertions:
+    count:
+      'div.bcl-content-banner': 1
+      'div.bcl-content-banner.bg-white': 1
+      'div.card > div.row > div.col-md-4.col-lg-3 > img[src="https://placeimg.com/1200/600/tech"]': 0
```
brummbar: There's no need to check for this exact image. We can broaden this up, and check for any `img`. They shouldn't be any.

templates/patterns/content_banner/content-banner.ui_patterns.yml
donquixote: Problem is we (currently) can't create these value objects in twig when including from another template, and we can't create them in yml for test fixtures.

So there are different options:
- technically support plain arrays instead of value objects, for pattern input. This is kind of what we currently do, we technically support it but we don't document it. So, "hidden feature, to be removed at any time".
- give up on yml test fixtures for these patterns, _and_ create the value objects in preprocess when we want to include the pattern in a template. This is what oe_theme does. Does not look ideal to me.
- Create twig functions to construct value objects. This still leaves the problem for yml test fixtures.
- Give up on value objects. Only applicable to new patterns, which this is.

I noticed in oe_theme we don't have yml fixtures for these patterns, and we always construct the value objects in preprocess, e.g. oe_theme_content_person_preprocess_oe_theme_helper_in_page_navigation__node__oe_person().

There are also some patterns in oe_theme where we don't use value objects.

Personally I wonder about the usefulness of these value objects in a twig / theming context. In a pure php context I absolutely support them. But I am sure this has been discussed and decided in the past. So..

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -0,0 +1,115 @@
...
+      meta:
+        - '26 May 2020'
+        - '1 comment'
+      title: 'This is the title of this page'
+      content: 'Lorem ipsum dolor sit amet, consectetur adipiscing elit. Suspendisse volutpat ultricies massa, a dapibus libero porta nec.'
+      badges:
+        - label: 'Primary badge'
+          rounded_pill: true
+        - label: 'Secondary badge'
+          rounded_pill: true
```
brummbar, line 101: We pass rounded_pill, but never an assertion that this is passed along?

brummbar (changes requested):

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -0,0 +1,43 @@
...
+    <div class="mb-4 mt-4">
+  {% endif %}
+  {% for _meta in meta %}
+    <span class="text-muted me-3">{{ _meta }}</span>
+  {% endfor %}
+  {% if meta is not empty %}
+    </div>
+  {% endif %}
+  <div>{{ content }}</div>
+{% endset %}
```
donquixote, line 18: We should wrap the entire block in if/else, not just the `<div>` and `</div>`.

templates/patterns/content_banner/pattern-content-banner.html.twig
donquixote: We need to / should check if the image is empty. The new version does that.
We can still use inline using ternary expression.

```twig
  'image': image ? {
    'path': image.src,
    'alt': image.alt,
  } : {},
```
I think this is more readable because we don't have to scroll around between where the variable is set and where it is used.

donquixote (approved):
Looks good now. All the recommended changes were applied. Thanks everybody!

brummbar (approved):

PR #177, opened 2022-02-07
OEL-1130: Remove table theming

drishu:
- Removed templates for table theming


escuriola (approved):

templates/overrides/dataset/table.html.twig
```diff
@@ -1,13 +0,0 @@
-{#
```
escuriola, line 1: This template can not be deleted. Must exists I suppose because dataset needs it and there is not one from system templates.

escuriola (changes requested):

templates/overrides/dataset/table.html.twig
donquixote: That's weird. There definitely is a `table.html.twig` in system module.
So what happens if we delete this one?

escuriola (dismissed):

templates/overrides/dataset/table.html.twig
escuriola: Yes, it was cache problem. Checked again and don't need to recover the pattern.

abel-santos-corral (dismissed):

escuriola (approved):

abel-santos-corral (approved):

PR #178, opened 2022-02-09
OEL-1159: Align search variant with default card pattern

donquixote:
Jira issue(s):
- [OEL-1159](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-1159)

### DIFF
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -16,7 +16,7 @@ card:
     title:
       type: text
       label: 'Title'
-      description: 'Card title.'
+      description: 'Card title. Supports markup, e.g. for links.'
       preview: 'Title card'
     subtitle:
       type: text
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -44,6 +44,15 @@ card:
       preview:
         path: 'https://placeimg.com/1000/500/people'
         alt: 'Alternative text for card image'
+    meta:
+      # Full type: `list<text|render|markup>`.
+      # Markup should be avoided if possible.
+      type: array
+      label: 'Meta items'
+      description: 'Date or other meta information.'
+      preview:
+        - 'Organisation'
+        - 'Position'
     content:
       # Full type: `string|render|markup`.
       # This is meant to be the most flexible parameter.
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -10,25 +10,34 @@ card:
       description: 'Card with special behaviour for print search results items.'
   fields:
     orientation:
+      # Full type: `'horizontal'|'vertical'`.
       type: string
       label:  'Orientation'
       description: 'Can be set to horizontal or vertical. Defaults to vertical.'
     title:
+      # Full type: `string|render|markup`.
+      # Markup should be avoided, except for links.
       type: text
       label: 'Title'
       description: 'Card title. Supports markup, e.g. for links.'
       preview: 'Title card'
     subtitle:
+      # Full type: `string|render|markup`.
+      # Markup should be avoided if possible.
       type: text
       label: 'Subtitle'
       description: 'Card subtitle.'
       preview: 'Subtitle card'
     text:
+      # Full type: `string|render|markup`.
+      # Markup can be used e.g. for <p> tags.
       type: text
       label: 'Text'
       description: 'Card text.'
       preview: 'Text card with longer text which is supported'
     image:
+      # Full type: `array{path: string, alt: string, ..}`.
+      # Look for '_image' in 'bcl-card.html.twig'.
       type: array
       label: 'Image'
       description: 'Card image with path, alt and position.'
@@ -36,21 +45,30 @@ card:
         path: 'https://placeimg.com/1000/500/people'
         alt: 'Alternative text for card image'
     content:
+      # Full type: `string|render|markup`.
+      # This is meant to be the most flexible parameter.
+      # Markup can be used for buttons or other elements.
       type: render
       label: 'Content'
       description: 'Card content.'
       preview: '<button class="btn btn-primary">Custom action</button>'
     header:
+      # Full type: `string|render|markup`.
+      # Markup should be avoided if possible.
       type: text
       label: 'Card header'
       description: 'Card header content.'
       preview: 'Header of card'
     footer:
+      # Full type: `string|render|markup`.
+      # Markup should be avoided if possible.
       type: text
       label: 'Card footer'
       description: 'Card footer content.'
       preview: 'Footer of card'
     badges:
+      # Full type: `list<array{label: string|render|markup, ..}>`.
+      # See 'bcl-badge.html.twig'.
       type: array
       label: 'Badges'
       description: 'List of badges for the card.'
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -36,8 +36,6 @@ card:
         path: 'https://placeimg.com/1000/500/people'
         alt: 'Alternative text for card image'
     meta:
-      # Full type: `list<text|render|markup>`.
-      # Markup should be avoided if possible.
       type: array
       label: 'Meta items'
       description: 'Date or other meta information.'
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -21,6 +21,13 @@ card:
       label: 'Title'
       description: 'Card title. Supports markup, e.g. for links.'
       preview: 'Title card'
+    url:
+      type: string
+      label: 'Url (optional)'
+      description: |
+        If present, the title is displayed as a link.
+        Only use this if the title is not already a link.
+      preview: 'https://www.example.com/'
     subtitle:
       # Full type: `string|render|markup`.
       # Markup should be avoided if possible.
--- a/templates/patterns/card/card.ui_patterns.yml
+++ b/templates/patterns/card/card.ui_patterns.yml
@@ -64,6 +64,8 @@ card:
       label: 'Badges'
       description: 'List of badges for the card.'
       preview:
-        - label: 'Primary badge'
-        - label: 'Secondary badge'
-          rounded_pill: true
+        - label: 'A badge'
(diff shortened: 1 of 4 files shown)


templates/patterns/card/card.ui_patterns.yml
```diff
@@ -10,47 +10,81 @@ card:
       description: 'Card with special behaviour for print search results items.'
   fields:
     orientation:
+      # Full type: `'horizontal'|'vertical'`.
```
escuriola: We don't have any comment. Please remove them.

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -10,47 +10,81 @@ card:
...
       description: 'Can be set to horizontal or vertical. Defaults to vertical.'
     title:
+      # Full type: `string|render|markup`.
+      # Markup should be avoided, except for links.
       type: text
       label: 'Title'
-      description: 'Card title.'
+      description: 'Card title. Supports markup, e.g. for links.'
       preview: 'Title card'
+    url:
```
escuriola: I'm not agree with the url field. We already have the title field. I don't feel comfortable with two fields for the same when we are trying to keep the patterns as simple as possible. 

Also If anybody needs a link then they maybe want to customise, a custom class, title, etc. And with url you loose this possibility you have with the title field. Maybe if instead url it was a link pattern and then print the link in the template, it would have more sense. But IMHO with title field is enough for cover this.

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -2,61 +2,73 @@
...
+ *
+ * Differences to main variant:
+ *   - Always use horizontal layout, with specific classes.
+ *   - Add extra classes on card body (right column), to prevent padding.
+ *     This should probably be handled by BCL if horizontal is used.
+ *   - Use text color dark.
+ *   - Set different margin-bottom for text element.
+ *   - Set special parameters on the image.
+ *   - Add special wrapper attributes.
+ *   - Ignore 'subtitle', 'date', 'card_header', 'card_footer'.
```
escuriola: Really need this? Please remove.

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -10,47 +10,81 @@ card:
...
       type: text
       label: 'Card footer'
       description: 'Card footer content.'
       preview: 'Footer of card'
     badges:
+      # Full type: `list<array{label: string|render|markup, ..}>`.
+      # See 'bcl-badge.html.twig'.
       type: array
       label: 'Badges'
       description: 'List of badges for the card.'
```
escuriola, line 63: The badges label are primary and secondary but has the same aspect, so I would change the label or add the badge style to the badges.

templates/patterns/card/pattern-card.html.twig
```diff
@@ -5,51 +5,52 @@
...
-  'title': title,
-  'subtitle': subtitle,
-  'text': text,
+    'tag': 'div',
+  } : {},
   'content': content,
   'image': image,
   'card_header': header,
   'card_footer': footer,
   'badges': badges,
```
escuriola: default badges has set the outline to true.

templates/patterns/card/pattern-card.html.twig
```diff
@@ -5,51 +5,52 @@
...
-  'date': _date,
+  'date': (date is not empty) ? {
+    'day': date|date('d'),
+    'month': date|date('M'),
+    'year': date|date('Y'),
+    'date_time': date|date('Y-m-d'),
+  } : {},
```
escuriola: We can remove date until full date ticket is prepared.

escuriola (changes requested):

templates/patterns/card/card.ui_patterns.yml
donquixote: I am ok to see it removed.
The motivation was that calling code should not have to set the 'underline-hover' class on the link.
But ok for me to remove.

templates/patterns/card/card.ui_patterns.yml
donquixote: I am confused by the wording of this comment.

templates/patterns/card/pattern-card.html.twig
donquixote: I notice that date is not even declared in the pattern definition.
I kept it there because it existed in the current version of the template.
But I agree to remove it.

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -35,6 +35,15 @@ card:
       preview:
         path: 'https://placeimg.com/1000/500/people'
         alt: 'Alternative text for card image'
+    meta:
+      # Full type: `list<text|render|markup>`.
```
escuriola: Please remove comments

escuriola (changes requested):

escuriola (dismissed):

tests/src/Kernel/fixtures/markup_rendering_patterns/card.yml
```diff
@@ -31,6 +34,8 @@ card_horizontal_with_image:
       'h5.card-title': 'Title card'
       'h6.card-subtitle': 'Subtitle card'
       'div.card-text': 'Text card with longer text which is supported'
+      '.card-body > div > span.text-muted:nth-child(1)': 'DIGIT'
```
brummbar, line 38: Can we add `.card-body > div > span.text-muted: 2` in the count assertions, so we make sure only 2 are rendered.

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -5,58 +5,54 @@
  */
 #}
 {% spaceless %}
```
brummbar: Let's take the chance to replace these with `{% apply spaceless %}` as this one used above is deprecated already.

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -5,58 +5,54 @@
...
+{% if badges is not empty %}
+  {% for item in badges %}
+    {% set _badges = _badges|merge([{
+      outline: true,
+      rounded_pill: true,
+    }|merge(item)]) %}
+  {% endfor %}
+{% endif %}
+{% include '@oe-bcl/card' with {
+  horizontal: true,
```
brummbar: In the other template we quote the keys, in here we don't. We should stick to one, I believe we went for always quote.

brummbar (changes requested):
Some minor remarks.

brummbar (dismissed):

escuriola (dismissed):

donquixote:
This will break whitelabel and showcase.
Let's not merge before the demo!

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -4,59 +4,55 @@
...
-      'border-md-0',
-      'border-0',
-      'mb-3',
-    ]}),
-    text_color: 'dark',
-    image: _image,
-    content: content
-  } only %}
-{% endspaceless %}
+  <div>{{ content }}</div>
```
drishu: is the compulsory div intentional ?

templates/patterns/card/pattern-card.html.twig
```diff
@@ -4,52 +4,51 @@
...
+{% set content %}
+  {% if meta is not empty %}
+    {# Use negative margin to compensate for me-3 on each item. #}
+    <div class="my-3 me-n3">
+      {% for _meta_item in meta %}
+        <span class="text-muted me-3">{{ _meta_item }}</span>
+      {% endfor %}
+    </div>
+  {% endif %}
+  <div>{{ content }}</div>
```
drishu: is the compulsory div intentional ?

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -4,59 +4,55 @@
...
+  },
+  'extra_classes_body': 'p-0 pb-md-0 pb-3',
+  'badges': _badges,
+  'title': {
+    'content': title,
+  },
+  'text': (text is not empty) ? {
+    'content': text,
+    'classes': 'mb-2-5',
+    'tag': 'div',
```
drishu, line 47: why div ? it was p before, I see p in netlify..

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -4,59 +4,55 @@
...
-          outline: true,
-          rounded_pill: true,
-          attributes: badge_attr
-        }]) %}
-      {% endif %}
-    {% endfor %}
+{% apply spaceless %}
+{% set content %}
+  {% if meta is not empty %}
+    {# Use negative margin to compensate for me-3 on each item. #}
```
drishu, line 10: lets remember to revisit this with BCL 0.21

drishu (changes requested):

templates/patterns/card/pattern-card.html.twig
donquixote: I think @brummbar  mentioned we should wrap this in a div. I don't have a strong opinion on it myself. 
In earlier versions of the theme, the idea was that "content" is like a general-purpose variable where developers can put additional stuff that is not already handled with other variables. so it would support more than one element, and these elements would already have their own wrappers.

This said: If we do keep the div, we need to check first that the content is not empty.

So we can either drop the div, or check emptiness first.

templates/patterns/card/pattern-card--variant-search.html.twig
donquixote: We want to support formatted text which may already contain `<p> `tags.
Then the `<div>` is just an additional wrapper.

templates/patterns/card/pattern-card--variant-search.html.twig
donquixote: see below.

templates/patterns/card/pattern-card--variant-search.html.twig
donquixote: ok but not now, right?

templates/patterns/card/card.ui_patterns.yml
donquixote: I think this is good now.

drishu (approved):

escuriola (approved):

PR #179, opened 2022-02-14
OEL-1178: Updates on status messages.

Maxfire:
Jira issue(s):
- OEL-1133: Changes needed to make the status message work out of the box.


Maxfire:
Done on OEL-1178

PR #191, opened 2022-02-24
OEL-755: Listing refactor & paragraph

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-755

Overrides: https://github.com/openeuropa/oe_bootstrap_theme/pull/128/files

### Prior versions pre-rebase
- #194 Original version from Carlos
- [First rebase, with separate commits for changes by Andreas](https://github.com/openeuropa/oe_bootstrap_theme/compare/976ff8ccc5168b16a0fd857a814ea47fe0cdc33d) (will be cleaned up by github garbage collection)
- current version has changes squashed.

PR #194, opened 2022-03-03
OEL-755: Backup

escuriola:
Backup for #191


escuriola:
Merged on OEL-755-resurrections

PR #195, opened 2022-03-07
OEL-1227: OEL-1227: Add content banner into a twig block.

GilNovacomm:
Jira issue(s):
- [OEL-1227](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1227)


donquixote:
Discarded approach.

PR #196, opened 2022-03-08
OEL-1160: Set navbar to default variant menu navigation.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1160

### DIFF
--- a/templates/overrides/menu/menu--main.html.twig
+++ b/templates/overrides/menu/menu--main.html.twig
@@ -4,6 +4,6 @@
  */
 #}
 {{ pattern('navigation', {
-  'variant': 'default',
+  'variant': 'navbar',
   'items': items
 }) }}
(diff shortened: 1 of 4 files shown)


brummbar (changes requested):
Let's add test assertions so that we check that navbar is on only for the default variant.

brummbar (dismissed):

brummbar (approved):

templates/overrides/menu/menu--main.html.twig
```diff
@@ -4,6 +4,6 @@
  */
 #}
 {{ pattern('navigation', {
-  'variant': 'default',
+  'variant': 'navbar',
```
drishu, line 7: We should set here class 'me-auto' and ditch the wrapping div in pattern-navbar.
If we do that and we override the block template for the main menu to remove the nav wrapper then we would have the needed structure in navbar. Let's do this in a follow up.

drishu (approved):

PR #201, opened 2022-03-11
[OEL-1255] OEL-1280: Remove most of paragraphs submodule (for move to oe_whitelabel)

abel-santos-corral:
Jira issue(s):
- [OEL-1280](https://citnet.tech.ec.europa.eu/CITnet/jira/projects/OEL/issues/OEL-1280)

Related PR in oe_whitelabel: [93](https://github.com/openeuropa/oe_whitelabel/pull/93)


modules/oe_bootstrap_theme_paragraphs/oe_bootstrap_theme_paragraphs.info.yml
```diff
@@ -1,21 +0,0 @@
-name: OpenEuropa Bootstrap Theme Paragraphs
```
drishu: you can't delete this file
```
[error]  OpenEuropa Bootstrap Theme Paragraphs is required if OpenEuropa Paragraphs is 
enabled.
 
 [error]   (Currently using Missing or invalid module The following module is marked as installed in the core.extension 
configuration, but it is missing:
```
just delete the config_devel from here

drishu (changes requested):
I will double check that all files are removed (test and whatever else related after the rebase, makes no sense now

drishu (changes requested):
Move also the paragraphs references in the README, and oe_bootstrap_theme_helper_requirements

drishu (changes requested):
found field--entity-reference-revisions--paragraph.html.twig

donquixote (approved):
Looks good to me now.

drishu (approved):

PR #209, opened 2022-03-17
OEL-1322: PHP 8.0 Upgrade.

escuriola:
Jira issue(s):
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1322

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -114,3 +114,4 @@ matrix:
     - highest
   PHP_VERSION:
     - 7.4
+    - 8.0
--- a/composer.json
+++ b/composer.json
@@ -22,6 +22,7 @@
         "drupal/drupal-extension": "^4.1",
         "drupal/file_link": "^2.0.6",
         "drupal/styleguide": "^1.0@beta",
+        "drupal/token": "^1.10",
         "drush/drush": "^10.3",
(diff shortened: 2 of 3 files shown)


.drone.yml
```diff
@@ -17,6 +17,7 @@ services:
     image: registry.fpfis.eu/fpfis/httpd-php:${PHP_VERSION}-ci
     environment:
       - DOCUMENT_ROOT=/test/oe_bootstrap_theme
+      - COMPOSERVER=--2
```
brummbar: We removed this already recently, no need to readd.

brummbar (changes requested):

brummbar (dismissed):

donquixote:
Seems good to me.

For the changes in composer.json, we need to make sure to clean up if we no longer need these packages. E.g. token is only needed for ui_patterns_settings, and perhaps we will remove this in the future?
But this should not block us.

I also don't know if there are any yet undiscovered problems for PHP 8.0.
But we can always create follow-up issues.

donquixote (dismissed):

donquixote:
Btw the README.md still talks about PHP 7.3 as a minimum version.
But I think we should deal with this in a separate issue, along with other README changes.

donquixote (approved):

brummbar (approved):

PR #211, opened 2022-03-22
OEL-1052: Enable media_avportal_mock on ParagraphsTestBase.

GilNovacomm:
Jira issue(s):
- [OEL-1052](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1052)

### DIFF
--- a/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
+++ b/tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
@@ -30,6 +30,7 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'locale',
     'media',
     'media_avportal',
+    'media_avportal_mock',
     'node',
     'oe_bootstrap_theme_paragraphs',
     'oe_media',


tests/src/Kernel/Paragraphs/ParagraphsTestBase.php
```diff
@@ -30,6 +30,7 @@ abstract class ParagraphsTestBase extends AbstractKernelTestBase {
     'locale',
     'media',
     'media_avportal',
+    'media_avportal_mock',
```
donquixote, line 33: Seems like a very simple fix, but quite hard to validate if I don't have the background in avportal.

But I did an experiment: Running the test with disconnected internet, with and without the fix.

MediaParagraphsTest::testFeaturedMedia - always fails without internet connection, because it has no access to oembed.com.

MediaParagraphsTest::testBanner:
- fails without the mock, with message "TypeError: Return value of Drupal\Core\File\FileUrlGenerator::doGenerateString() must be of the type string, null returned"
- passes with the mock, because it no longer needs to access the internet.

So this means the mock prevents any problems we could have from the avportal service not working.
This seems to be the goal here, so +1.
It does not tell us why the avportal service failed.

-------

This also reveals a flaw in avportal:
- Drupal\Core\StreamWrapper\StreamWrapperInterface::getExternalUrl() has a soft contract to return `string` (I say "soft" because it is only in the doc `@return` type, not a native return type). I checked other implementations, they don't return NULL.
- AvPortalPhotoStreamWrapper::getExternalUrl() breaks this contract by returning NULL if the client returns an empty (NULL) resource. I don't know what the "correct" behavior would be: Throw an exception? The contract on StreamWrapperInterface does not say.
- FileUrlGenerator::doGenerateString() has a native return type and _must_ return string. But instead it returns NULL because of the misbehaving stream wrapper.

In fact in file_create_url() there is a try/catch, and it catches `InvalidStreamWrapperException`.
So this could be the correct exception to be thrown.
Of course this would only shift the problem to a different place, because now we would have FALSE for the image src in oe_bootstrap_theme_preprocess_paragraph__oe_banner().

brummbar (approved):

escuriola (approved):

PR #216, opened 2022-03-25
OEL-494: File pattern.

brummbar:


patches/npm/@openeuropa+bcl-theme-default+0.20.0.patch
```diff
@@ -0,0 +1,13 @@
+diff --git a/node_modules/@openeuropa/bcl-theme-default/templates/bcl-file/file.html.twig b/node_modules/@openeuropa/bcl-theme-default/templates/bcl-file/file.html.twig
+index 66ea86a..465cd4f 100644
+--- a/node_modules/@openeuropa/bcl-theme-default/templates/bcl-file/file.html.twig
++++ b/node_modules/@openeuropa/bcl-theme-default/templates/bcl-file/file.html.twig
+@@ -168,7 +168,7 @@
+               {% if _item.title is not empty %}
+                 <p class="fw-bold m-0">{{ _item.title }}</p>
+               {% endif %}
+-              {% if _item.title is not empty %}
++              {% if _item.description is not empty %}
```
escuriola, line 10: is reporter this error to FRONT?

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -34,6 +49,16 @@ public function __construct(CacheableDependencyInterface $plugin_manager_block)
   public function getFilters(): array {
     return [
       new TwigFilter('bcl_card_list', [$this, 'bclCardList']),
+      new TwigFilter('format_size', 'format_size'),
```
drishu, line 52:
```suggestion
      new TwigFilter('format_size', 'formatSize'),
```

templates/patterns/file/pattern-file.html.twig
```diff
@@ -0,0 +1,74 @@
...
+    },
+    items: _items,
+  } %}
+{% endif %}
+
+{% set _meta = '(' ~ file.extension|upper  ~ ')' %}
+{% if file.size %}
+  {% set _meta = '(' ~ file.size|format_size ~ ' - ' ~ file.extension|upper  ~ ')' %}
+{% endif %}
+{% endapply %}
```
drishu, line 47: no consistency in core as usual but we could apply spaceless to the entire template

tests/src/Kernel/MarkupRenderingTest.php
```diff
@@ -206,6 +220,30 @@ protected function assertMarkupRendering(array $assertions, string $html): void
     }
   }
 
+  /**
+   * Executes the pattern assertion class specified in the test data.
+   *
+   * @param array $assertions
+   *   An associative array with the following keys:
+   *   - class: A pattern assertion QCN.
```
drishu: what's QCN?

tests/src/Kernel/fixtures/PatternTestDataMassager.php
```diff
@@ -0,0 +1,75 @@
...
+
+declare(strict_types = 1);
+
+namespace Drupal\Tests\oe_bootstrap_theme\Kernel\fixtures;
+
+use Drupal\oe_bootstrap_theme\ValueObject\FileValueObject;
+
+/**
+ * Transforms test data coming from YAML files.
+ *
```
drishu, line 11: I think this is a good candidate for internal use

src/ValueObject/FileValueObject.php
```diff
@@ -0,0 +1,271 @@
...
+
+  /**
+   * Getter.
+   *
+   * @return string
+   *   Property value.
+   */
+  public function getExtension(): string {
+    if ($this->extension === NULL) {
+      $mime_types = new MimeTypes();
```
drishu, line 220: lets create a ticket, or leave a todo, to analyses solutions for figuring out the mime of remote files (a nice to have for future dev)

templates/patterns/file/pattern-file.html.twig
```diff
@@ -0,0 +1,74 @@
...
+      language: _item.language_code|to_native_language,
+      meta: _translation_meta,
+      download: {
+        label: link_label|default('Download'|t),
+        path: _item.url,
+      },
+    }]) %}
+  {% endfor %}
+  {% set _translations = {
+    label: {
```
drishu, line 29: One edge case I found is that if you have a document printed twice on the page and you click to open the translations both (or all n instances) of this printed on the page will open at the same time. 
This is because the id of the translations collapse is cached in drupal.

tests/src/PatternAssertion/FilePatternAssert.php
```diff
@@ -0,0 +1,142 @@
...
+   * @param array $expected
+   *   The expected values.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertItem(array $expected, Crawler $crawler): void {
+    $this->assertElementExists(sprintf('a[href="%s"]', $expected['url']), $crawler);
+    $this->assertElementText($expected['title'], 'p.fw-bold.m-0', $crawler);
+
+    $data_container = $crawler->filter('p.fw-bold.m-0 + small.fw-bold');
```
drishu, line 129: dono how this passed here, but in whitelabel when we use it, as soon as we add translations this fails. This is because `p.fw-bold.m-0 + small.fw-bold` fetches the first small after the mentioned p, but the translations have a `<p class="m-0"></p>` in between so $data_container is null.
```suggestion
    $data_container = $crawler->filter('small.fw-bold');
```

drishu (changes requested):

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
brummbar: format_size is a Drupal function, see https://api.drupal.org/api/drupal/core%21includes%21common.inc/function/format_size/9.3.x -

templates/patterns/file/pattern-file.html.twig
brummbar: We should limit the use of spaceless where we have multiple blank lines due to our processing, not everywhere by default. In this case I do exactly that.

tests/src/Kernel/MarkupRenderingTest.php
brummbar: Fully Qualified Class name without Fully 😄

tests/src/Kernel/fixtures/PatternTestDataMassager.php
brummbar: It's in tests, no need to mark it.

src/ValueObject/FileValueObject.php
brummbar: It's already done by file_link module.

templates/patterns/file/pattern-file.html.twig
brummbar: It's a bug on the script and on our side. We split it off and handle it later.

tests/src/PatternAssertion/FilePatternAssert.php
brummbar: Wrote it as comment in the ticket: on whitelabel the patch is not applied because we are not using a release, so there it's mean to fail until we make a temporary release.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
drishu: sry was a reflex

templates/patterns/file/pattern-file.html.twig
drishu: I understand, agreed

tests/src/Kernel/MarkupRenderingTest.php
drishu: I should know this, thx

tests/src/Kernel/fixtures/PatternTestDataMassager.php
drishu: test classes like traits can also be used externally (we also do), but if we don't mind the usage then its ok for me

drishu (dismissed):

drishu (approved):

escuriola (approved):

PR #222, opened 2022-03-29
OEL-1380: Paragraph field bare template suggestion should not depend on oe_bootstrap_theme_paragraphs

### LINKED ISSUE #221: OEL-1380: Paragraph field bare template suggestion should not depend on oe_bootstrap_theme_paragraphs
## Problem

In 9a44fe1d6b4a410667406f1a41ec98e155bbde43, the `field__bare` suggestion was moved from the theme to `oe_bootstrap_theme_paragraphs` sub-module. The `oe_bootstrap_theme_paragraphs` sub-module is designed to bridge between `oe_bootstrap_theme` and `oe_paragraphs` module. But this is not the case for all projects. Some projects, such as Joinup, are using the `paragraphs` module without depending on `oe_paragraphs`. The change from 9a44fe1d6b4a410667406f1a41ec98e155bbde43 did introduce a regression when we updated OE Bootstrap Theme from an early version to `1.0.0-alpha7`.

## Proposal

Move back (revert) the suggestion alter to `includes/field.inc` so that all projects can benefit for this template suggestion, even they are not using `oe_paragraphs`.

claudiu-cristea:
See [OEL-1380](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1380). Fixes #221

### DIFF
--- a/includes/field.inc
+++ b/includes/field.inc
@@ -29,6 +29,14 @@ function oe_bootstrap_theme_preprocess_field(&$variables) {
 function oe_bootstrap_theme_theme_suggestions_field_alter(array &$suggestions, array $variables): void {
   $element = $variables['element'];
 
+  // Do not output field labels and wrapping markup for paragraph fields.
+  if (isset($element['#entity_type']) && $element['#entity_type'] === 'paragraph') {
+    // Prepend the new suggestion to the list. This will put it right after the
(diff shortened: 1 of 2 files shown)


donquixote:
won't do here.
Paragraphs are moving to oe_whitelabel.

PR #229, opened 2022-04-07
OEL-1425: Change image arrays to ImageValueObject in patterns.

escuriola:
Jira issue(s):
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1425


brummbar (approved):

drishu (approved):
I already approve, up to you and @brummbar , we could also do banner, it is used in whitelabel paragraphs and the preprocess already instantiates an ImageValueObject

brummbar:
We are using it in banner paragraph but wrong, as we only use the src at the end. So it's not needed.

PR #230, opened 2022-04-13
OEL-1416: Up BCL 0.20.0 -> 0.22.0

donquixote:
Jira issue(s):
- [OEL-1416](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-1416)


tests/src/Kernel/fixtures/markup_rendering_patterns/date_block.yml
```diff
@@ -6,10 +6,13 @@ date_block_default:
...
-      'time[datetime="2021-05-26"]': 1
-      'span.py-4.bg-white.border.border-bottom-0.w-100.fs-5.fw-medium.text-dark.text-center.text-uppercase': 1
-      'span.bg-lighter.border.w-100.text-center.fs-5.text-dark.rounded-bottom.mb-n1': 1
+      'time': 1
+      'time.bcl-date-block[datetime="2021-05-26"]': 1
+      'time > *': 1
+      'time > * > *': 2
+      'time > * > * > *': 1
+      'time > * > * > * > *': 0
+      'time > * > *:nth-child(2) > *': 0
```
donquixote: These counts verify there are no unexpected elements.
TBD.

tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml
```diff
@@ -42,12 +42,12 @@ pagination_with_icon:
...
       prev:
         path: '#prev'
+        # Use a custom icon for 'prev'.
         icon:
           name: 'arrow-left'
       next:
         path: '#next'
-        icon:
-          name: 'arrow-right'
+        # Use the built-in icon for 'next'.
```
donquixote, line 50: This is a temporary solution until we do OEL-1463.

tests/src/Kernel/fixtures/markup_rendering_patterns/date_block.yml
drishu: honestly I think we where checking to many classes previously but now it looks to cryptic, can u amend this to I recognize the date structure, or at least leave a comment as to what I'm seeing here

drishu (changes requested):

drishu (dismissed):

tests/src/Kernel/fixtures/markup_rendering_patterns/date_block.yml
```diff
@@ -6,10 +6,13 @@ date_block_default:
...
-      'time[datetime="2021-05-26"]': 1
-      'span.py-4.bg-white.border.border-bottom-0.w-100.fs-5.fw-medium.text-dark.text-center.text-uppercase': 1
-      'span.bg-lighter.border.w-100.text-center.fs-5.text-dark.rounded-bottom.mb-n1': 1
+      # Exclude unexpected additional <time> elements.
+      'time': 1
+      'time.bcl-date-block[datetime="2021-05-26"]': 1
+      # Exclude unexpected nested elements, no matter the tag name.
+      'time > *': 1
+      'time > * > *': 2
+      'time > * > *:nth-child(1) > *': 1
```
brummbar: I don't see the value for this, the most important assertions to run here are checking that the day and month are in the first box and year is in the second. Let's drop these altogether.

brummbar (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/date_block.yml
donquixote: Ok for me to remove.

However it makes me wonder, when do we ever want to have these count checks?
Obviously their purpose is to fail if we have unexpected element.

I could see two different reasons why we would not want to do this:
- We allow BCL to insert additional elements without having to update our tests.
- We trust that unexpected additional elements are unlikely to appear, so it is not worth the additional lines of testing code.

donquixote: Let's keep the `'time.bcl-date-block[datetime="2021-05-26"]': 1` for the datetime attribute.

brummbar (approved):

drishu (approved):

PR #232, opened 2022-04-21
OEL-1360: Clean button for modals close button.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1360


drishu (approved):

drishu (dismissed):
I think the spaceless should be with apply filter and should encompass also the set _id, but not a blocker

brummbar (dismissed):

drishu (approved):

brummbar (approved):

PR #234, opened 2022-04-22
OEL-1463: Pager first and last, PaginationPatternAssert

donquixote:
[OEL-1463](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1463)


templates/overrides/views/views-mini-pager.html.twig
```diff
@@ -5,14 +5,19 @@
...
+      'path': items.previous.href,
+    },
+    'items': [{
+      'label': items.current,
+      'disabled': true,
+    }],
+    'next': items.next ? {
+      'label': items.next.text,
+      'path': items.next.href,
+    },
```
donquixote, line 21: I wanted to add first/last links here, but then realized that we don't do that for views-mini-pager.
However, I still wanted to clean up the unnecessary local variables.

templates/patterns/pagination/pagination.ui_patterns.yml
```diff
@@ -18,11 +18,22 @@ pagination:
         center: 'center'
         end: 'end'
       preview: 'center'
-    enable_icons:
+    enable_prev_next_icon:
```
drishu: why split?

tests/src/Kernel/PagerTest.php
```diff
@@ -0,0 +1,264 @@
...
+          'links' => $links,
+        ],
+        $page,
+      ];
+    }
+    return $argss;
+  }
+
+  /**
+   * Tests a common pager on different pages.
```
drishu: something with views though

tests/src/Kernel/PagerTest.php
```diff
@@ -0,0 +1,264 @@
...
+   * Renders a render element, but does not expect a by-reference parameter.
+   *
+   * @param array $element
+   *   Element to render.
+   *
+   * @return string
+   *   The rendered HTML.
+   */
+  protected function renderByValue(array $element): string {
+    return $this->renderRoot($element);
```
drishu: what's the point of having a wrapper around this function ?

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -0,0 +1,110 @@
...
+      'a' => $expected_count,
+      'nav > ul > li.page-item > a.page-link:first-child' => $expected_count,
+    ], $crawler);
+  }
+
+  /**
+   * Asserts the alignment.
+   *
+   * @param string $expected
+   *   The expected icon size.
```
drishu: 'The expected icon size.' I see this in a lot of places, is it really the icon size that is expected everywhere ?

drishu (changes requested):
Please also check a view with full pager, first last don't appear

templates/patterns/pagination/pagination.ui_patterns.yml
donquixote: BCL has two distinct parameters to enable first/last and prev/next icons.
By default only one of them is enabled.
With this parameter split, we replicate the signature and variability of BCL.

tests/src/Kernel/PagerTest.php
donquixote: It allows inline calls, because it is not by reference.
```php
$element = [...];
render($element);
```
vs
```php
render([...]);
```
But you are free to disagree, maybe it is overkill now. In another version I had more calls to this method.

donquixote: Removed.

templates/patterns/pagination/pagination.ui_patterns.yml
drishu: now I see that in the designs its mixed, so in the end ok for me

donquixote: Actually there can be usability reasons to show text for prev/next but icons for first/last.
- prev/next may be clicked more often than first/last, and the text makes the clickable area bigger.
- having text for one thing and icons for another thing makes them more easily distinguishable by shape.

drishu (dismissed):

drishu (dismissed):

tests/src/Kernel/PagerTest.php
```diff
@@ -0,0 +1,256 @@
...
+  }
+
+  /**
+   * Data provider.
+   *
+   * @return array[]
+   *   Argument lists.
+   */
+  public function pagerDataProvider(): array {
+    $linkss = [
```
brummbar: We can rename this `$scenarios`.

brummbar: We can rename this `$scenarios`.

tests/src/Kernel/PagerTest.php
```diff
@@ -0,0 +1,256 @@
...
+      ],
+      7 => [
+        ['url' => '?page=0', 'icon' => 'chevron-double-left'],
+        ['url' => '?page=6', 'label' => 'Previous'],
+        ['url' => '?page=5', 'label' => '6'],
+        ['url' => '?page=6', 'label' => '7'],
+        ['url' => '?page=7', 'label' => '8', 'active' => TRUE],
+      ],
+    ];
+    $argss = [];
```
brummbar: Let's drop the double s, maybe we can call this tests or data.

brummbar: Let's drop the double s, maybe we can call this tests or data.

tests/src/Kernel/PagerTest.php
```diff
@@ -0,0 +1,256 @@
...
+  public function testCustomLabels(): void {
+    \Drupal::request()->query->set('page', '2');
+
+    /** @var \Drupal\Core\Pager\PagerManagerInterface $pager_manager */
+    $pager_manager = $this->container->get('pager.manager');
+    $pager_manager->createPager(75, 10);
+
+    $element = [
+      '#type' => 'pager',
+      '#quantity' => 3,
```
brummbar, line 42: We are always testing with quantity = 3, we should have one test with no quantity param specified.

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -0,0 +1,115 @@
...
+   *     - 'icon': The name of the icon.
+   *     - 'active': Whether the item has an 'active' class.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertLinks(array $expected, Crawler $crawler): void {
+    $expected_bcl_icon_url = '/' . \Drupal::service('extension.list.theme')->getPath('oe_bootstrap_theme') . '/assets/icons/bcl-default-icons.svg';
+
+    $actual_links_data = array_map(
+      function (\DOMElement $actual_li_element) use ($expected_bcl_icon_url) {
```
brummbar: Let's have a simple loop here.

brummbar: Let's have a simple loop here.

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -0,0 +1,115 @@
...
+      iterator_to_array($crawler->filter('nav > ul > li'), FALSE),
+    );
+
+    $this->assertSame($expected, $actual_links_data);
+
+    $expected_count = count($expected);
+    $this->assertCounts([
+      'li' => $expected_count,
+      'a' => $expected_count,
+      'nav > ul > li.page-item > a.page-link:first-child' => $expected_count,
```
brummbar, line 62: Is this to assert that links are first?  Just asking, not a remark.

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -0,0 +1,115 @@
...
+      function (\DOMElement $actual_li_element) use ($expected_bcl_icon_url) {
+        $li = new Crawler($actual_li_element);
+        $link = $li->filter('a');
+        $actual_link_data = [];
+        $actual_link_data['url'] = $link->attr('href');
+        $use = $li->filter('a > svg > use');
+        if ($use->count()) {
+          // This is an icon.
+          $icon_url = $use->attr('xlink:href');
+          [$actual_bcl_icon_url, $actual_link_data['icon']] = explode('#', $icon_url, 2) + [NULL];
```
brummbar: This code is hard to read, especially the extra [NULL]. Let's try to make these things easier to understand by dropping a comment.

brummbar: This code is hard to read, especially the extra [NULL]. Let's try to make these things easier to understand by dropping a comment.

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -0,0 +1,115 @@
...
+          $this->assertSame($expected_bcl_icon_url, $actual_bcl_icon_url);
```
brummbar: You are making a big assertSame down to have a full picture on the errors, but then run a loose assert here.
We need to have a common way of running assertions and comparisons.
For example we have the icon assert, and we could and should use that one here.

We need to choose one approach, fail early or full asserts.

brummbar: You are making a big assertSame down to have a full picture on the errors, but then run a loose assert here.
We need to have a common way of running assertions and comparisons.
For example we have the icon assert, and we could and should use that one here.

We need to choose one approach, fail early or full asserts. Let's stick to fail early like in file pattern assert, so we can reuse existing assertions inside.

brummbar (changes requested):

tests/src/Kernel/PagerTest.php
donquixote: `$datasets` is how these things are called in phpunit.

tests/src/PatternAssertion/PaginationPatternAssert.php
donquixote: My strategy with count asserts is like this:
- First count with the broadest selector possible, to catch all `<a>` or `<li>` tags, including those in unexpected places or with missing classes.
- Then count again with strictest selectors for position/structure and classes/attributes.

If both counts are the same, we know that all existing `<a>` or `<li>` tags are in the expected position and have the expected attributes.

The `:first-child` makes sure that every `<li>` has exactly one `<a>`. E.g. it would fail if we have one empty `<li></li>` and then one `<li><a></a><a></a></li>` with more than one link tag.

"My bag contains exactly 3 items of clothing and 15 buttons, _and_ my bag contains 3 yellow shirts size M each having 5 buttons." -> no loose leftover buttons, no other items of clothing, no other shirts with different color or size.

tests/src/PatternAssertion/PaginationPatternAssert.php
donquixote: If we want what you call "full asserts", I think we need to get there gradually. So we would have to endure a situation or a phase where we have both.

My motivation here was to cover the most common fails with a group assert that provides detailed information for fixing the test, and then have additional single assertions for things that will fail less often, and that are harder to cover in the group assertion.

The group assert failure output was already helpful in fixing the datasets.

This said, I agree for consistency it may be better to go for single asserts all the way, even if that makes the failure output less helpful.

donquixote: duplicate comment, resolving :)

tests/src/PatternAssertion/PaginationPatternAssert.php
donquixote: duplicate, resolving

donquixote: Actually the `+ [NULL]` is not just hard to read, it is wrong / insufficient.
The goal is to always have an array with two elements to avoid "Warning: Undefined array key 1".
To achieve this we need `+ [NULL, NULL]`, not just `+ [NULL]`.
https://3v4l.org/aATpQ

Much better would be if PHP would allow a default value in the array unpacking.
https://3v4l.org/JRgC2
But this is not the case.

I am going to drop the array unpacking and use separate assignments instead.
(and then drop the entire thing to go for single assert)

tests/src/PatternAssertion/PaginationPatternAssert.php
donquixote: Yep.

Btw, PhpStorm asks me to add `ext-dom` to composer.json.
I won't, because it is already required by other packages, and I think we don't do this elsewhere.

brummbar (dismissed):

drishu (dismissed):
love the more straight forward approach in the pattern assert

templates/patterns/pagination/pagination.ui_patterns.yml
```diff
@@ -18,11 +18,22 @@ pagination:
         center: 'center'
         end: 'end'
       preview: 'center'
-    enable_icons:
```
escuriola, line 21: We can take advantage and remove settings section from pattern definition

templates/patterns/pagination/pagination.ui_patterns.yml
escuriola:
```suggestion
    enable_prev_next_icons:
```
It's a suggestion

templates/patterns/pagination/pagination.ui_patterns.yml
```diff
@@ -18,11 +18,22 @@ pagination:
...
+    enable_prev_next_icon:
       type: boolean
-      label: 'Enable icon'
+      label: 'Enable prev/next icon'
       description: 'Use icon instead of link labels to previous and next items. False by default.'
+    enable_first_last_icon:
```
escuriola: Same

templates/patterns/pagination/pattern-pagination.html.twig
```diff
@@ -8,9 +8,12 @@
   'size': size,
   'alignment': alignment,
   'aria_label': aria_label,
```
escuriola, line 10: We can remove aria_label. Can be added via attributes

escuriola (changes requested):

templates/patterns/pagination/pattern-pagination.html.twig
donquixote: We can, but should we?
I am already overtime in this issue, and this change is not in original scope.

templates/patterns/pagination/pagination.ui_patterns.yml
donquixote: I took the parameter name from BCL.
But I like your suggestion!

templates/patterns/pagination/pagination.ui_patterns.yml
donquixote: Let's do this in a new ticket for all patterns.

drishu (approved):

escuriola (approved):

brummbar (approved):

PR #235, opened 2022-04-25
OEL-1442: Fix pager prev next default values.

GilNovacomm:
Jira issue(s):
- [OEL-1442](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1442)


brummbar:
Fixed in #234 .

PR #239, opened 2022-05-09
OEL-1383: Fix slimSelect, only activate when a multiple select exists.

escuriola:
Jira issue(s):
- https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1383


brummbar (dismissed):
I've been wondering if, since we are just adding one class in the template, we should go for something like
```
{% set attributes = attributes.addClass([
  attributes.hasClass('error') ? 'is-invalid',
]) %}
{% include "@system/select.html.twig" %}
```
But it seems that in core they went for full copying templates so let's kepe it this way for now.

drishu (dismissed):

brummbar (approved):

drishu (approved):

PR #241, opened 2022-05-18
OEL-1382: Allow drupal alters in menu links.

escuriola:
Jira issue(s):
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1382

### DIFF
--- a/includes/menu.inc
+++ b/includes/menu.inc
@@ -7,24 +7,50 @@
 
 declare(strict_types = 1);
 
+use Drupal\Core\Link;
 use Drupal\Core\Template\Attribute;
+use Drupal\Core\Url;
 
 /**
  * Implements hook_preprocess_menu().
  */
-function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
+function oe_bootstrap_theme_preprocess_menu(array &$variables, string $hook): void {
   if ($hook === 'menu__toolbar') {
     return;
   }
 
+  // Merge additional attributes into the URL object.
+  $prepare_url = function (array $item): Url {
+    // Don't modify the original URL object.
+    $url = clone $item['url'];
+
+    // Merge passed attributes into the URL object.
+    $passed_attributes = [];
+    if (!empty($item['attributes'])) {
+      $passed_attributes = $item['attributes'] instanceof Attribute ? $item['attributes']->toArray() : $item['attributes'];
+    }
+    $attributes = [
+      'class' => $item['in_active_trail'] ? ['active'] : [],
+    ] + $passed_attributes;
+
+    // This is a parent item.
+    if (!empty($item['below'])) {
+      $attributes['class'][] = 'nav-link';
+    }
+
+    $url->mergeOptions(['attributes' => $attributes]);
+
+    return $url;
+  };
+
   $variables['pattern_items'] = [];
   foreach ($variables['items'] as $key => $item) {
+    $url = $prepare_url($item);
     $pattern_item = [
       'label' => $item['title'],
-      'path' => $item['url']->toString(),
-      'attributes' => new Attribute([
-        'class' => $item['in_active_trail'] ? ['active'] : [],
-      ]),
+      'url' => $item['url'],
+      'attributes' => new Attribute($url->getOption('attributes')),
+      'link' => (new Link($item['title'], $item['url']))->toRenderable(),
     ];
     if (empty($item['below'])) {
       $variables['pattern_items'][$key] = $pattern_item;
@@ -32,7 +58,6 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
     }
 
     $submenu = $pattern_item;
-    $submenu['attributes'] = new Attribute(['class' => 'nav-link']);
     $pattern_item += [
       'standalone' => TRUE,
       'dropdown' => TRUE,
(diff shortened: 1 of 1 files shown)


includes/menu.inc
```diff
@@ -7,24 +7,46 @@
...
+
+    return $url;
+  };
+
   $variables['pattern_items'] = [];
   foreach ($variables['items'] as $key => $item) {
+    $url = $prepare_url($item);
     $pattern_item = [
       'label' => $item['title'],
+      'url' => $item['url'],
```
claudiu-cristea: This needs to be `'url' => $url`, I guess

includes/menu.inc
```diff
@@ -7,24 +7,46 @@
...
+    $url = clone $item['url'];
+
+    // Merge passed attributes into the URL object.
+    $passed_attributes = [];
+    if (!empty($item['attributes'])) {
+      $passed_attributes = $item['attributes'] instanceof Attribute ? $item['attributes']->toArray() : $item['attributes'];
+    }
+    $attributes = [
+      'class' => $item['in_active_trail'] ? ['active'] : [],
+    ] + $passed_attributes;
```
claudiu-cristea: What if `$passed_attributes` already contains a `class` attribute? Wouldn't that be overridden here?

includes/menu.inc
```diff
@@ -41,17 +63,17 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
       'trigger' => $submenu,
     ];
     foreach ($item['below'] as $key_sub_item => $sub_item) {
+      $url = $prepare_url($sub_item);
       $pattern_item['items'][$key_sub_item] = [
         'label' => $sub_item['title'],
+        'url' => $sub_item['url'],
```
claudiu-cristea: Same. I think we should assign `$url` because it might be different.

includes/menu.inc
```diff
@@ -7,24 +7,46 @@
...
+  };
+
   $variables['pattern_items'] = [];
   foreach ($variables['items'] as $key => $item) {
+    $url = $prepare_url($item);
     $pattern_item = [
       'label' => $item['title'],
+      'url' => $item['url'],
+      'attributes' => new Attribute($url->getOption('attributes')),
+      'link' => (new Link($item['title'], $item['url']))->toRenderable(),
```
claudiu-cristea: Should pass `$url` rather than `$item['url']`.

includes/menu.inc
```diff
@@ -41,17 +63,17 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
       'trigger' => $submenu,
     ];
     foreach ($item['below'] as $key_sub_item => $sub_item) {
+      $url = $prepare_url($sub_item);
       $pattern_item['items'][$key_sub_item] = [
         'label' => $sub_item['title'],
+        'url' => $sub_item['url'],
+        'attributes' => new Attribute($url->getOption('attributes')),
+        'link' => (new Link($sub_item['title'], $sub_item['url']))->toRenderable(),
```
claudiu-cristea: Should pass `$url` rather than `$item['url']`.

claudiu-cristea (changes requested):

claudiu-cristea:
Certainly, this needs tests

includes/menu.inc
```diff
@@ -7,51 +7,70 @@
...
+      $submenu = $pattern_item;
+      $pattern_item += [
+        'standalone' => TRUE,
+        'dropdown' => TRUE,
+        'link' => TRUE,
+        'alignment' => 'dropdown-menu_item-sm-start',
+        'trigger' => $submenu,
       ];
+      foreach ($item['below'] as $key_sub_item => $sub_item) {
+        $pattern_item['items'][$key_sub_item] = $get_pattern_item[$sub_item];
```
drishu, line 69:
```
$get_pattern_item($sub_item);
```
we cant use anonymous function as an array because they are implemented as Closure classes

includes/menu.inc
```diff
@@ -7,51 +7,70 @@
...
+    /** @var \Drupal\Core\Url $url */
+    $url = clone $item['url'];
+
+    // Merge passed attributes into the URL object.
+    $attributes = [];
+    if (!empty($item['attributes'])) {
+      $attributes = $item['attributes'] instanceof Attribute ? $item['attributes']->toArray() : $item['attributes'];
+    }
+    $class = $attributes['class'] ?? [];
+    $class[] = 'nav-link';
```
drishu, line 34: like this submenu items receive the nav-link class which throws off the design of it,
its ok to remove it, the nav-link class is re-attached in the navigation bcl to the correct items.

includes/menu.inc
```diff
@@ -7,51 +7,70 @@
...
+    if ($item['in_active_trail']) {
+      $class[] = 'active';
     }
+    $attributes['class'] = $class;
+    $url->mergeOptions(['attributes' => $attributes]);
 
-    $submenu = $pattern_item;
-    $submenu['attributes'] = new Attribute(['class' => 'nav-link']);
```
drishu, line 35: and then if we remove 'nav-link' from prepare_url then we have to keep this

includes/menu.inc
```diff
@@ -7,51 +7,70 @@
...
+
+  $variables['pattern_items'] = [];
+  foreach ($variables['items'] as $key => $item) {
+    $pattern_item = $get_pattern_item($item);
+    if (!empty($item['below'])) {
+      $submenu = $pattern_item;
+      $pattern_item += [
+        'standalone' => TRUE,
+        'dropdown' => TRUE,
+        'link' => TRUE,
```
drishu, line 64: this is what we found in netlify at the time, don't know if this needs to be set and TRUE (or something that equivalates to true, like the object set in prepare_url), but with the new setup it does not seem needed anymore here.
I did look into it and could not find any usage of the item.link in the bcl templates.

drishu (changes requested):
Personally i agree with the idea of not having a template for it, its just that there is some logic in the template that makes it look the way BCL needs it, we will have to move that also into the preprocess

drishu (commented):
We will follow up with a different direction, see ticket

PR #242, opened 2022-05-19
OEL-1618: Update BCL to 0.23.0

escuriola:
Jira issue(s):
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1618

### DIFF
--- a/templates/patterns/card/pattern-card.html.twig
+++ b/templates/patterns/card/pattern-card.html.twig
@@ -38,7 +38,6 @@
     gutter: 0,
   } : {},
   'title': title,
-  'title_tag': 'h5',
   'subtitle': (subtitle is not empty) ? {
     'content': subtitle,
     'classes': 'mb-2',
--- a/templates/patterns/card/pattern-card.html.twig
+++ b/templates/patterns/card/pattern-card.html.twig
@@ -37,9 +37,8 @@
     right_col_classes: 'col-8',
     gutter: 0,
   } : {},
-  'title': (title is not empty) ? {
-    'content': title,
-  } : {},
+  'title': title,
+  'title_tag': 'h5',
   'subtitle': (subtitle is not empty) ? {
     'content': subtitle,
     'classes': 'mb-2',
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -31,7 +31,6 @@
 {% include '@oe-bcl/bcl-content-banner/bcl-content-banner.html.twig' with {
   'background': background,
   'title': title,
-  'title_tag': 'h2',
   'content': _content,
   'image': image ? {
     'path': image.src,
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -30,9 +30,8 @@
 
 {% include '@oe-bcl/bcl-content-banner/bcl-content-banner.html.twig' with {
   'background': background,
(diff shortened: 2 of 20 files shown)


templates/patterns/card/pattern-card.html.twig
```diff
@@ -37,9 +37,8 @@
     right_col_classes: 'col-8',
     gutter: 0,
   } : {},
-  'title': (title is not empty) ? {
-    'content': title,
-  } : {},
+  'title': title,
+  'title_tag': 'h5',
```
donquixote: I talked with front-end.
The idea is that card titles should now be h1. Semantically this is ok because they are inside `<article>`, so they only apply to the article.
This means we don't need to set it to 'h5'.

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -30,9 +30,8 @@
 
 {% include '@oe-bcl/bcl-content-banner/bcl-content-banner.html.twig' with {
   'background': background,
-  'title': {
-    'content': title,
-  },
+  'title': title,
+  'title_tag': 'h2',
```
donquixote: Title tag should remain `h1` especially for content banner, because now this is the page title.

tests/src/Kernel/fixtures/markup_rendering_patterns/progress.yml
```diff
@@ -37,7 +37,7 @@ progress_secondary_non_stripped_non_animated:
     '#id': progress
     '#fields':
       settings:
-        style: 'secondary'
+        style: 'light'
```
donquixote, line 40: What's the goal here? You want to test that the text color changes if you switch to 'light'?
And we no longer need to test 'secondary'?

I would say:
- If we still want to test 'secondary' we should instead add another scenario to test with 'light'.
- If we want to drop the 'secondary' scenario and only test 'light', we should rename the scenario. E.g. 'progress_light_non_stripped_non_animated'.

I think there is probably nothing special about 'secondary', so probably the second option is ok.

templates/patterns/offcanvas/pattern-offcanvas--variant-search.html.twig
```diff
@@ -7,11 +7,8 @@
 {% extends 'pattern-offcanvas.html.twig' %}
 {% block offcanvas %}
   {% include '@oe-bcl/offcanvas' with {
-    'title': {
-      'label': title,
-      'heading': 4,
-      'extra_classes': "mb-lg-4",
-    },
+    'title': title,
+    'title_attributes': create_attribute().addClass('mb-lg-4'),
```
donquixote, line 11: So both variants will now use h3 title tag. Ok for me.

donquixote (dismissed):
This one seems ok. The other prs still need work.

donquixote:
I think we need a follow-up for featured media paragraph + pattern.
The pattern template is weird because it contains so much custom html instead of using bcl.

The heading in the pattern is `h3`. To align with other paragraphs it should be `h2`.
In netlify it is `h2.fw-bold`, which is probably a mistake: https://oelibrary.netlify.app/default/?path=/story/paragraphs-featured-media--featured-item

EDIT: Actually it is ok on the "Landing page" where the paragraph is used: https://oelibrary.netlify.app/default/?path=/story/features-landing-page--default

Let's fix here, no need for separate issue.

donquixote (dismissed):

brummbar (dismissed):

drishu (approved):
on behalf of @donquixote

brummbar (approved):

PR #245, opened 2022-05-23
OEL-1540: Do not load inpage_navigation pattern if there are no links to display

escuriola:
https://citnet.tech.ec.europa.eu/CITnet/jira/browse/OEL-1540

### DIFF
--- a/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
+++ b/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
@@ -6,11 +6,12 @@
 #}
 {% if links is not null %}
   {{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
-  {% include '@oe-bcl/inpage-navigation' with {
-    'title': title,
-    'links': links,
-    'content': content,
-    'full_layout': true,
-    'id': inpage_navigation_id,
-  } only  %}
 {% endif %}
+{% include '@oe-bcl/inpage-navigation' with {
+  'title': title,
+  'links': links,
+  'content': content,
+  'full_layout': true,
+  'id': inpage_navigation_id,
+} only  %}
+
--- a/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
+++ b/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
@@ -4,7 +4,7 @@
  * Inpage navigation pattern.
  */
 #}
-{% if links is not null %}
+{% if links is not empty %}
   {{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
 {% endif %}
 {% include '@oe-bcl/inpage-navigation' with {
--- a/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
+++ b/templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
@@ -4,12 +4,13 @@
  * Inpage navigation pattern.
  */
 #}
-{{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
-
-{% include '@oe-bcl/inpage-navigation' with {
-  'title': title,
-  'links': links,
-  'content': content,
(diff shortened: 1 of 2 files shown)


drishu (dismissed):

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -4,12 +4,13 @@
...
+{% if links is not null %}
+  {{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
+  {% include '@oe-bcl/inpage-navigation' with {
+    'title': title,
+    'links': links,
+    'content': content,
+    'full_layout': true,
+    'id': inpage_navigation_id,
+  } only  %}
+{% endif %}
```
donquixote, line 9: This solution has some problems:
- If `links` is `null`, but `content` is non-empty, then the content won't be printed. To fix this, we would have to make an `else` block. This can happen e.g. if the content does not contain any headers for scrollspy. E.g. in the oe_content_row paragraph type where we use this, all the items could have empty titles.
- Are we sure `links` will be null? What if it is an empty array?
- Do we want to show the inpage nav if there is only one link?

Also we may ask:
- Should it be the pattern's responsibility to provide a fallback behavior if no links, or should the calling template do this, or both? In other words, what is the "contract" for the pattern? E.g. we could say it is forbidden to call the pattern if links is empty.

We should also consider if we can test these special cases, e.g.
- (whitelabel) content row paragraphs with no titles
- (whitelabel) events where we would have no inpage nav but still have some content to show
- (oebt) inpage nav pattern with content but without nav links

All of this should be discussed with the team.

templates/patterns/inpage_navigation/pattern-inpage-navigation.html.twig
```diff
@@ -4,12 +4,14 @@
  * Inpage navigation pattern.
  */
 #}
-{{ attach_library('oe_bootstrap_theme/inpage_navigation') }}
-
+{% if links is not null %}
```
donquixote: @escuriola As mentioned, please check if links not empty, not just not null.
We don't want to show them for empty array.

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -75,3 +75,22 @@ inpage_navigation_non_full:
...
+      content:
+        '#markup': '<div class="content_example">The content is rendering but there are no links</div>'
+  assertions:
+    count:
+      'nav.position-sticky': 1
+      'nav[id^="bcl-inpage-navigation"]': 1
+      'h5': 1
+      'ul.nav.nav-pills.flex-column': 1
+      'li.nav-item': 0
+      'a.nav-link': 0
```
donquixote, line 93: Ok for me. It is silly to have an inpage nav with zero links, but this is the consequence of our decision.
The test describes the existing behavior, and if we change it, we have to change the test.

The test does _not_ cover that javascript is not included. It would be interesting to check that, but I don't know if it is possible with these fixture based tests.

I leave it for next reviewer (QA) to agree with this or not.

donquixote (approved):

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -75,3 +75,22 @@ inpage_navigation_non_full:
       'a[href="#item-3"]': 'Heading 3'
       'a[href="#item-4"]': 'Heading 4'
       'h5': 'Page content'
+inpage_navigation_with_no_links:
```
drishu, line 78: the there is not '#attached' or 'script' available to assert against, so this will have to do

drishu (dismissed):

donquixote (dismissed):

donquixote (approved):

donquixote (approved):

drishu (approved):

PR #258, opened 2022-08-01
OEL-1900: Upgrade BCL to 0.25.

drishu:
Jira issue(s):

### DIFF
--- a/tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
+++ b/tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
@@ -27,7 +27,7 @@ content_banner_white_background_with_image_with_badges_with_meta_with_links:
       'div.bcl-content-banner': 1
       'div.bcl-content-banner.bg-white': 1
       'img': 1
-      'article.card > div.row > div.col-md-4.col-lg-3 > img[src="https://placeimg.com/1200/600/tech"]': 1
+      'article.card > div.row > div.bcl-card-start-col > img[src="https://placeimg.com/1200/600/tech"]': 1
       'div.card-body > h1': 1
       'div.card-body > div': 4
       'div.card-body > div.mt-4': 0
@@ -87,7 +87,7 @@ content_banner_gray_background_with_image_without_badges_without_meta_without_li
       'div.bcl-content-banner': 1
       'div.bcl-content-banner.bg-lighter': 1
       'img': 1
-      'article.card > div.row > div.col-md-4.col-lg-3 > img[src="https://placeimg.com/1200/600/tech"]': 1
+      'article.card > div.row > div.bcl-card-start-col > img[src="https://placeimg.com/1200/600/tech"]': 1
       'div.card-body > h1': 1
       'div.card-body > div.mt-4': 1
       'div.card-body > div > span.badge': 0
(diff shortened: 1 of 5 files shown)


tests/src/Kernel/fixtures/markup_rendering_patterns/description_list.yml
```diff
@@ -35,31 +35,30 @@ description_list_default:
...
-      'svg.me-2.bi.icon--xl': 1
-      'dl > dt:nth-of-type(5) > svg.me-2.bi.icon--xl': 1
+      'svg.bi.icon--xs': 2
+      'svg.bi.icon--xl': 1
+      'dl:nth-of-type(4) > dt > svg.bi.icon--xl': 1
     contains:
-      'dl > dt:nth-of-type(1) > svg': '/themes/custom/oe_bootstrap_theme/assets/icons/bcl-default-icons.svg#geo-alt-fill'
-      'dl > dt:nth-of-type(3) > svg': '/themes/custom/oe_bootstrap_theme/assets/icons/bcl-default-icons.svg#calendar-event'
-      'dl > dt:nth-of-type(5) > svg': '/themes/custom/oe_bootstrap_theme/assets/icons/bcl-default-icons.svg#bicycle'
+      'dl:nth-of-type(1) > dt svg': '/themes/custom/oe_bootstrap_theme/assets/icons/bcl-default-icons.svg#geo-alt-fill'
```
brummbar, line 46: The fact that now we have multiple dl elements instead of one for me is a regression/backwards compatibility break and we need to address this.

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -87,7 +87,7 @@ content_banner_gray_background_with_image_without_badges_without_meta_without_li
       'div.bcl-content-banner': 1
       'div.bcl-content-banner.bg-lighter': 1
       'img': 1
-      'article.card > div.row > div.col-md-4.col-lg-3 > img[src="https://placeimg.com/1200/600/tech"]': 1
+      'article.card > div.row > div.bcl-card-start-col > img[src="https://placeimg.com/1200/600/tech"]': 1
```
brummbar, line 90: Is this change in size expected?
Before
image

and now
image

It doesn't match the UX files I found.

brummbar (changes requested):
To discuss with UX/FE on some points.

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
emeryro: I had a look in figma, and indeed on a content page the image is supposed to be large.
I don't know on which content type/page this is, but that's probably a mistake
image

planctus: This is the consequence of the card refactoring, a parameter is now in place to determine the image size, you can see here  in the content banner demo how you can set the image size:
https://oelibrary.netlify.app/default/?path=/story/compositions-content-banner--default&args=image_size:xl

tibi2303: In the second image from francesco is the default size (160x160) which is displayed on the listings pages - https://oelibrary.netlify.app/default/?path=/story/features-person--listing
and the large size (255x255) is displayed on the details pages (content pages) 
https://oelibrary.netlify.app/default/?path=/story/features-person--details

As davide mentioned, we introduced 4 sizes which can be manipulated with the image_size param inside content_banner :
- 120*120 - sm
- 160*160 - md (default)
- 255*255 - lg
- 350*350 - xl - no use cases for it now so it does not have a demo

PR #265, opened 2022-08-19
OEL-1912: Revamp carousel pattern.

drishu:


templates/patterns/carousel/carousel.ui_patterns.yml
```diff
@@ -1,54 +1,19 @@
...
-      type: boolean
-      label: 'Show controls'
-      description: 'Enable the controls (prev and next buttons). False by default.'
-      preview: true
-    show_indicators:
-      type: boolean
-      label: 'Show indicators'
-      description: 'With indicators (default: false).'
-      preview: true
-    autoplay:
```
drishu: I feel we could keep this one, what do you think ?

templates/patterns/carousel/pattern-carousel.html.twig
```diff
@@ -4,30 +4,12 @@
...
 {% include '@oe-bcl/carousel' with {
   'id': carousel_id,
-  'autoplay': autoplay,
-  'autoinit': autoinit,
-  'dark': dark_mode,
-  'active_item': active_item,
-  'items': _slides,
-  'fade': fade,
-  'with_controls': show_controls,
+  'with_controls': true,
```
drishu: I know there are defaults, but want to set in stone the behaviour

templates/patterns/carousel/carousel.ui_patterns.yml
```diff
@@ -1,54 +1,19 @@
 carousel:
   label: 'Carousel'
   description: 'A slideshow component for cycling through elements, like a carousel. For more information: https://v5.getbootstrap.com/docs/5.0/components/carousel/'
-  settings:
```
brummbar: Unfortunately we shipped with these field already, so we cannot remove them anymore. I didn't realise for some reason that we had already a pattern when I did the analysis. We need to leave it untouched in this version.

tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -0,0 +1,71 @@
...
+   * Asserts the carousel pattern items.
+   *
+   * @param array[] $expected
+   *   The expected item values.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertItems(array $expected, Crawler $crawler): void {
+    $items = $crawler->filter('.carousel .carousel-inner .carousel-item');
+    self::assertCount(count($expected), $items);
```
brummbar:
```suggestion
    self::assertSameSize($expected, $items);
```

brummbar (changes requested):

drishu (commented):
.

tests/src/Kernel/fixtures/markup_rendering_patterns/carousel.yml
```diff
@@ -8,90 +8,206 @@ carousel_with_fade_show_controls_with_indicator_without_disable_touch_without_au
...
+            caption: 'First slide caption'
+            caption_classes: 'caption_classes-1'
+            interval: 1
+          - image:
+              src: 'https://images.site.example2'
+              alt: 'alt image 2'
+            caption_title: 'Second slide title'
+            caption: 'Second slide caption'
+            caption_classes: 'caption_classes-2'
+            interval: 2
```
drishu, line 222: theming this I see now I need to add "link"

tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -0,0 +1,148 @@
...
+  protected function assertSettings(array $expected, Crawler $crawler): void {
+    if (!empty($expected['fade'])) {
+      $this->assertElementExists('.carousel-fade', $crawler);
+    }
+    else {
+      $this->assertElementNotExists('.carousel-fade', $crawler);
+    }
+
+    if (!empty($expected['show_controls'])) {
+      $this->assertElementText('Previous', '.carousel-control-prev', $crawler);
```
brummbar: Most of these methods can receive NULL and they will check for absence of element/attribute.
We could rewrite (some of) them like this:
```suggestion
      $this->assertElementText($expected['show_controls'] ?? NULL, '.carousel-control-prev', $crawler);
```

tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -0,0 +1,148 @@
...
+
+  /**
+   * Asserts the carousel pattern settings.
+   *
+   * @param array[] $expected
+   *   The expected settings.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertSettings(array $expected, Crawler $crawler): void {
```
brummbar, line 45: I think we should provide the default values for all the settings and always test for them, as we pass them directly down to the template and to form the "api layer" we need to make sure their values don't change.

tests/src/PatternAssertion/BasePatternAssert.php
```diff
@@ -125,6 +127,30 @@ protected function assertElementText($expected, string $selector, Crawler $crawl
...
+   * Asserts text is contained in a particular element.
+   *
+   * @param string|null $expected
+   *   The expected value.
+   * @param string $selector
+   *   The CSS selector to find the element.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertElementTextContains(?string $expected, string $selector, Crawler $crawler): void {
```
brummbar, line 140: I guess this is in line with other assertions, just leaving this as a note. We changed the element attribute assertion to check if the attribute is present or not, so maybe this one too should check if the element has the text or not. For now I think it's fine to leave it like it is.

brummbar (changes requested):

tests/src/PatternAssertion/BasePatternAssert.php
drishu: Not sure, current implementation still permits sending an empty string in $expected, but my main concern about this is that in our tests cases the actual antagonist of this if a missing element, not an empty value of the attribute

brummbar: True, we can use empty string. Perfect!

brummbar (dismissed):

brummbar (approved):

PR #267, opened 2022-09-01
OEL-2166: Upgrade to BCL 1.0.0

brummbar:


kits/default/package.json
```diff
@@ -13,7 +13,7 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "0.24.1",
+    "@openeuropa/bcl-builder": "0.25.1",
```
drishu: Lets use the same 28.1

docker-compose.yml
```diff
@@ -28,7 +28,7 @@ services:
     user: "node"
     working_dir: /home/node/app
     environment:
-      - NODE_ENV=${NODE_ENV}
+      - NODE_ENV=${NODE_ENV:-''}
```
drishu: Can you give me a bit on info, the commit message should also be a bit more descriptive. 
`:-` evaluates to use `''` if NODE_ENV is unset or empty, but we give an empty string as default, would `-` (use '' if NODE_ENV is unset) work ?

drishu (changes requested):
I also noticed inpage navigation is not working.
And while testing in showcase I noticed the offcanvas bit in the search form is visible only in mobile.

docker-compose.yml
brummbar: We'll replace with `-`  only as it's the same operation done by Docker: provides an empty string as fallback only when the variable is not defined at all.

brummbar:
Fixed the offcanvas, for which I reviewed the updates back in the days but forgot to bring them. The inpage navigation is broken so I'll create an issue for it.

drishu (approved):

PR #270, opened 2022-09-12
OEL-2002: Preserve Drupal links.

drishu:
Jira issue(s):


includes/menu.inc
```diff
@@ -17,41 +17,26 @@ function oe_bootstrap_theme_preprocess_menu(&$variables, $hook) {
...
-  $variables['pattern_items'] = [];
-  foreach ($variables['items'] as $key => $item) {
-    $pattern_item = [
-      'label' => $item['title'],
-      'path' => $item['url']->toString(),
-      'attributes' => new Attribute([
-        'class' => $item['in_active_trail'] ? ['active'] : [],
-      ]),
-    ];
+  $menu_preprocess = \Drupal::classResolver(MenuPreprocess::class);
```
drishu, line 17: I dont think this is actually needed with this class, we could just instantiate with new

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -235,4 +251,33 @@ public function toInternalLanguageId(string $language_code): string {
...
+   *   The URL object or string used for the link.
+   * @param array|\Drupal\Core\Template\Attribute $attributes
+   *   An optional array or Attribute object of link attributes.
+   *
+   * @return array
+   *   The link render array.
+   */
+  public function bclLink(Environment $env, $label, $path, Attribute $attributes): array {
+    if (empty($path)) {
+      $path = 'route:<nolink>';
```
drishu, line 271: I found this is how core handles it

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -235,4 +251,33 @@ public function toInternalLanguageId(string $language_code): string {
...
+    }
+
+    // Here we handle internal urls which already had toString() called.
+    if (is_string($path) && !UrlHelper::isExternal($path)) {
+      $path = str_replace(base_path(), '/', $path);
```
drishu, line 276: didn't find anything in core, closes it something in FileUrlGenerator but its not the same.

templates/patterns/navigation/navigation.ui_patterns.yml
```diff
@@ -33,7 +33,7 @@ navigation:
           path: '#third'
         - label: "I'm a link active"
           active: true
-          target: '#fourth'
+          path: '#fourth'
```
drishu, line 36: nothing like "target" is actually handled

patches/npm/@openeuropa+bcl-theme-default+0.24.1.patch
```diff
@@ -11,6 +11,50 @@ index 6e08be9..ddc41ad 100644
...
+->
+-  {%- if _icon_position == 'before' and _icon_markup is defined -%}
+-    {{ _icon_markup|raw }}
+-  {%- endif -%}
+-  {{- _label -}}
+-  {%- if _icon_position == 'after' and _icon_markup is defined -%}
+-    {{ _icon_markup|raw }}
+-  {%- endif -%}
+-</a>
++{% if _icon_markup is defined %}
```
drishu, line 41: In the end this is how I handle it, bclLink receives a Twig\Markup object and till then label is handled as is. (Didn't have success trying to handle this in a render array build)

drishu (commented):
One more thing, I want to add a new test for it, or change the link pattern render test where we cover all possible scenarios of urls, anchors, external links, full internal urls, uris, url objects whatever else.

patches/npm/@openeuropa+bcl-theme-default+0.24.1.patch
```diff
@@ -11,6 +11,50 @@ index 6e08be9..ddc41ad 100644
...
++  {%- set _label -%}
++    {%- if _icon_position == 'before'  -%}
++      {{ _icon_markup|raw }}
```
brummbar: Is the |raw needed? I tried locally and it seems it wasn't. When we set rendered html to a variable in a twig file, it gets set as safe markup.

templates/overrides/menu/menu-local-tasks.html.twig
```diff
@@ -5,11 +5,15 @@
  */
 #}
 {% if primary %}
-  <h2 class="visually-hidden">{{ 'Primary tabs'|t }}</h2>
```
brummbar, line 8: A bit sad we lose this :(

templates/patterns/icon/pattern-icon.html.twig
```diff
@@ -7,5 +7,6 @@
 {% include '@oe-bcl/icon' with {
   'name': name,
   'path': bcl_icon_path,
-  'size': size
+  'size': size,
+  'attributes': attributes,
```
brummbar, line 11: Should we define this in the pattern declaration?

tests/src/PatternAssertion/LinkPatternAssert.php
```diff
@@ -0,0 +1,125 @@
...
+      $this->assertElementNotHasClass('standalone', 'a', $crawler);
+    }
+
+    if (isset($expected['icon_position'])) {
+      $link = $crawler->filter('a')->getNode(0);
+
+      // If no icon or no label is set then the position is irrelevant.
+      if (count($link->childNodes) === 2) {
+        $node_name = $expected['icon_position'] === 'before' ? 'svg' : '#text';
+        self::assertSame($node_name, $link->firstChild->nodeName);
```
brummbar: There's a scenario that now I forgot, which doesn't make me like assertSame for scalar assertions. I remember only that it's harder to debug the errors with assert same, I'd rather use assertEquals.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -235,4 +251,52 @@ public function toInternalLanguageId(string $language_code): string {
...
+   * @param \Drupal\Core\Url|string $path
+   *   The URL object or string used for the link.
+   * @param array|\Drupal\Core\Template\Attribute $attributes
+   *   An optional array or Attribute object of link attributes.
+   *
+   * @return array
+   *   The link render array.
+   */
+  public function bclLink(Environment $env, $label, $path, Attribute $attributes): array {
+    if (is_string($path) && !UrlHelper::isExternal($path)) {
```
brummbar: By looking at breadcrumb template/preprocess, they use directly markup to render the link. So we should go with the following: if the path is already a string, we return a rendered `<a>` element. We can use an inline template or return a render array 
```
return [
  '#type' => 'html_tag',
  '#tag'=> 'a',
  '#value' => $label,
  '#attributes' => $attributes,
];
```

brummbar: We should do the check in here because checking if a variable is string is not possible out of the box in twig.

brummbar (changes requested):

templates/overrides/menu/menu-local-tasks.html.twig
drishu: Can't remember why I removed it, so I put it back :|

templates/patterns/icon/pattern-icon.html.twig
drishu: Yes, I amended the whole icon track

tests/src/PatternAssertion/LinkPatternAssert.php
drishu: ok, I changed also in the TwigExtension test

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -235,4 +252,37 @@ public function toInternalLanguageId(string $language_code): string {
...
+   *   An optional array or Attribute object of link attributes.
+   *
+   * @return array
+   *   The link render array.
+   */
+  public function bclLink(Environment $env, $label, $path, Attribute $attributes): array {
+    if (is_string($path)) {
+      // The text has been processed by twig already, convert it to a safe
+      // object for the render system.
+      if ($label instanceof TwigMarkup) {
```
drishu, line 274: had to do this else we get to HtmlTag.phpL105 where the Xss filter removes our icon, so if it came from Twig, I deemed it safe just as in the cores TwigExtension::getLink

modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
```diff
@@ -0,0 +1,178 @@
...
+        ],
+        [
+          'path' => '/',
+          'label' => 'Custom attrib link',
+          'attributes' => [
+            'data-foo' => 'baz',
+            'id' => 'id-bar',
+          ],
+        ],
+      ],
```
brummbar, line 174: I think we miss a scenario with external url.

tests/src/PatternAssertion/BasePatternAssert.php
```diff
@@ -114,6 +114,53 @@ protected function assertElementAttribute($expected, string $selector, string $a
...
+   * @param string $selector
+   *   The CSS selector to find the element.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertElementHasClass($expected, string $selector, Crawler $crawler): void {
+    $this->assertElementExists($selector, $crawler);
+    $class = $crawler->filter($selector)->attr('class');
+
+    self::assertNotNull($class, \sprintf(
```
brummbar: Nit: no backslashes.

tests/src/PatternAssertion/LinkPatternAssert.php
```diff
@@ -0,0 +1,125 @@
...
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
+    $crawler = new Crawler($html);
+
+    $this->assertElementExists('a', $crawler);
```
brummbar: We should probably do 'body > a' to check that it's first level.

tests/src/PatternAssertion/LinkPatternAssert.php
```diff
@@ -0,0 +1,125 @@
...
+  /**
+   * Asserts the link icon.
+   *
+   * @param string $expected
+   *   The expected settings.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertIcon(string $expected, Crawler $crawler): void {
+    $svg = $crawler->filter('svg');
```
brummbar: We should `body > a > svg` to check that it's rendered inside.

tests/src/PatternAssertion/LinkPatternAssert.php
```diff
@@ -0,0 +1,125 @@
...
+   * Asserts the link pattern settings.
+   *
+   * @param array[] $expected
+   *   The expected settings.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertSettings(array $expected, Crawler $crawler): void {
+    if (!empty($expected['disabled'])) {
+      $this->assertElementHasClass('disabled', 'a', $crawler);
```
brummbar: Actually we don't need this method at all. Since we know there's only one `a` tag, we can simply assert that `a.disabled` exists.

tests/src/PatternAssertion/LinkPatternAssert.php
```diff
@@ -0,0 +1,125 @@
...
+      $this->assertElementAttribute('true', 'a', 'aria-disabled', $crawler);
+    }
+    else {
+      $this->assertElementNotHasClass('disabled', 'a', $crawler);
```
brummbar: This the same as above, but asserting that doesn't exist.

src/MenuPreprocess.php
```diff
@@ -0,0 +1,208 @@
...
+   * @param array $local_tasks
+   *   The local tasks.
+   *
+   * @return array
+   *   A list of filtered and processed links.
+   */
+  public function prepareLocalTasks(array $local_tasks): array {
+    $links = [];
+    foreach ($local_tasks as $link) {
+      if ($link['#access']->isForbidden()) {
```
brummbar: Core handles this differently: see build/core/lib/Drupal/Core/Render/Renderer.php:226.
Basically access is granted only if access is set to allowed, or to true.
In our case we will always have objects (\Drupal\Core\Menu\LocalTaskManager::getTasksBuild()).

tests/src/Kernel/MenuLocalTasksTest.php
```diff
@@ -74,6 +83,17 @@ public function testMenuLocalTasks(): void {
...
+        ],
+        'link7.link' => [
+          '#theme' => 'menu_local_task',
+          '#link' => [
+            'title' => 'Forbidden link',
+            'url' => Url::fromUri('http://www.forbiddenlink.com'),
+          ],
+          '#active' => FALSE,
+          '#weight' => 0,
+          '#access' => $access_forbidden,
```
brummbar, line 98: Let's add a case where access is neutral. Also in that case the access should be prevented.

brummbar (changes requested):

brummbar (approved):

PR #272, opened 2022-09-15
OEL-1989: Add outline parameter to badges.

drishu:
Jira issue(s):

### DIFF
--- a/tests/src/Kernel/fixtures/markup_rendering_patterns/badge.yml
+++ b/tests/src/Kernel/fixtures/markup_rendering_patterns/badge.yml
@@ -23,6 +23,26 @@ badge_with_all_values:
         label: 'A simple badge!'
         url: 'https://example.com'
         title: 'The badge title'
+badge_no_settings_sent:
+  render:
+    '#type': pattern
+    '#id': badge
+    '#fields':
+      label: 'A simple badge!'
+      url: 'https://example.com'
+      title: 'The badge title'
+  assertions:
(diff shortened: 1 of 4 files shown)


tests/src/Kernel/fixtures/markup_rendering_patterns/badge.yml
```diff
@@ -1,3 +1,28 @@
+badge_with_all_values:
```
brummbar, line 1: We have no test coverage for the case where we don't pass settings. Since we don't have default either in the pattern template, this can lead to regressions if BCL changes the values. Let's add a case without any settings.

tests/src/PatternAssertion/BadgePatternAssert.php
```diff
@@ -0,0 +1,81 @@
...
+   * @param array[] $expected
+   *   The expected settings.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The crawler.
+   */
+  protected function assertSettings(array $expected, Crawler $crawler): void {
+    $background = $expected['background'] ?? 'primary';
+
+    if (!empty($expected['outline'])) {
+      $this->assertElementExists('.badge-outline-' . $background, $crawler);
```
brummbar, line 60: Let's add here a negative assert for the .bg-$background case.

brummbar (changes requested):

brummbar (approved):

PR #274, opened 2022-09-20
OEL-2038: Patch BCL - Content banner sizes on different cases.

julien-:
Jira issue(s):
- [OEL-2038](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-2038)


julien-:
Making patch according [BCL PR 455](https://github.com/openeuropa/bootstrap-component-library/pull/455/files)

PR #286, opened 2022-11-24
ISAICP-7433: The #access property might be a bool

### LINKED ISSUE #285: Error when link #access property is a boolean
## Problem

In `MenuPreprocess::prepareLocalTasks()` a check to link `#access` assumes that this property is always an instance of `AccessResultInstance`. But this isn't always the case, link elements, as all render/form elements, accept also a _boolean_ as value for `#access`. Some core examples:

* https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/layout_builder/src/Form/LayoutBuilderEntityViewDisplayForm.php#L67
* https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/layout_builder/src/Element/LayoutBuilder.php#L374
* https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/layout_builder/src/Controller/ChooseBlockController.php#L122
* https://git.drupalcode.org/project/drupal/-/blob/9.5.x/core/modules/comment/src/Form/CommentAdminOverview.php#L218
* ...and so on

When a link having `#access` set as boolean is processed in that method, this error is triggered:

```
Error: Call to a member function isAllowed() on bool in Drupal\oe_bootstrap_theme\MenuPreprocess->prepareLocalTasks() (line 196 of themes/contrib/oe_bootstrap_theme/src/MenuPreprocess.php).
```

## Proposal

Accept also _booleans_ as values for `#access`.

claudiu-cristea:
Jira issue(s):
- [ISAICP-7433](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ISAICP-7433)

Fixes #285

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -14,6 +14,7 @@
         "openeuropa/ecl-twig-loader": "^3.1"
     },
     "require-dev": {
+        "ext-dom": "*",
         "composer/installers": "^1.11",
         "drupal/config_devel": "~1.8",
         "drupal/core-composer-scaffold": "^9.3",
--- a/src/MenuPreprocess.php
+++ b/src/MenuPreprocess.php
@@ -194,6 +194,7 @@ class MenuPreprocess {
   public function prepareLocalTasks(array $local_tasks): array {
     $links = [];
     foreach ($local_tasks as $link) {
+      // @see \Drupal\Core\Render\Element::isVisibleElement()
       $access = (!isset($link['#access'])
         || (($link['#access'] instanceof AccessResultInterface && $link['#access']->isAllowed()) || ($link['#access'] === TRUE)));
       if (!$access) {
--- a/src/MenuPreprocess.php
+++ b/src/MenuPreprocess.php
@@ -193,7 +193,8 @@ class MenuPreprocess {
   public function prepareLocalTasks(array $local_tasks): array {
     $links = [];
     foreach ($local_tasks as $link) {
-      if (!$link['#access']->isAllowed()) {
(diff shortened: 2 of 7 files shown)


claudiu-cristea:
See https://github.com/openeuropa/oe_bootstrap_theme/issues/285#issuecomment-1327746480

claudiu-cristea:
But, I think, the access check should be completely removed. The access is checked upstream, in renderer. Note that none of the menu local tasks preprocessor are checking the access, that's because the backend, not the theme layer, is responsible to perform this check. See:

* `olivero_preprocess_menu_local_tasks()`
* `claro_preprocess_menu_local_tasks()`

My advice is to remove this check.

drishu (commented):
I agree in principle with what you say @claudiu-cristea. 
Here is what I found, if we look at LocalTasksBlock::build() we see that Element::getVisibleChildren is used to determine if the tabs should be rendered, but if it passed then all elements in the tabs are passed along:
`'#primary' => count(Element::getVisibleChildren($links['tabs'])) > 1 ? $links['tabs'] : [],`

That being said I was expecting this to be filtered further down the line, so I improved the test to use the LocalTasksBlock, result is that the Links get rendered even the ones with access forbidden or false 
https://drone.fpfis.eu/openeuropa/oe_bootstrap_theme/2261/36

drishu (commented):
I've amended the local task preprocess to acomodate both boolean and the situation where #access is not set (in which case I see in Element::isVisibleElement that it should pass), please check both commits

drishu (approved):

PR #287, opened 2022-12-09
OEL-2128: Add Skip to main content target.

drishu:


templates/overrides/page/page.html.twig
```diff
@@ -60,6 +60,8 @@
       </header>
     {% endif %}
 
+    <a id="main-content" tabindex="-1"></a>
```
drishu, line 63: I've added this hear as it is after the header but not inside the page.content if, because I don't want it to be missing even if there is no content.

brummbar (approved):

PR #290, opened 2022-12-13
OEL-2009: Style tables using BCL.

brummbar:


modules/oe_bootstrap_theme_helper/tests/src/Unit/BootstrapTableFilterTest.php
```diff
@@ -0,0 +1,111 @@
...
+   * @param string $html
+   *   The filter input.
+   * @param string $expected
+   *   The expected filter output.
+   *
+   * @dataProvider processDataProvider
+   */
+  public function testProcess(array $settings, string $html, string $expected): void {
+    $filter = new BootstrapTable(['settings' => $settings], 'oe_bootstrap_theme_table', ['provider' => 'test']);
+    $filter->setStringTranslation($this->getStringTranslationStub());
```
drishu: This is not needed

templates/overrides/dataset/table.html.twig
```diff
@@ -0,0 +1,45 @@
...
+ *   any rows.
+ * - no_striping: A boolean indicating that the row should receive no striping.
+ * - header_columns: The number of columns in the header.
+ *
+ * @see template_preprocess_table()
+ *
+ * @ingroup themeable
+ */
+#}
+{% set attributes = attributes.addClass('table') %}
```
drishu: we discussed a theme wide setting

drishu (changes requested):

tests/src/Kernel/fixtures/markup_rendering.yml
```diff
@@ -199,3 +199,50 @@ status_messages_multiple:
...
+    count:
+      'table': 1
+      'table.table': 1
+      'td': 4
+table_bootstrap_responsive:
+  render:
+    '#theme': 'table'
+    '#rows':
+      - ['A', 'B']
+      - ['C', 'D']
```
drishu: this setup seems the same as the first one

drishu (changes requested):

drishu (approved):

PR #291, opened 2022-12-19
OEL-1915: Support date in card pattern.

drishu:


tests/src/PatternAssertion/CardPatternAssert.php
```diff
@@ -0,0 +1,197 @@
...
+   *   The expected data.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertDate(array $expected, Crawler $crawler): void {
+    // @todo Use dedicated pattern assert once we re-work date_plock pattern.
+    $day_month = $expected['day'] . ' ' . $expected['month'];
+
+    if (isset($expected['end_day']) && isset($expected['end_month'])) {
+      $day_month .= ' - ' . $expected['end_day'] . '      ' . $expected['end_month'];
```
drishu: I don't want to do extra processing here nor patch the template just for this strange looking extra space

tests/src/PatternAssertion/CardPatternAssert.php
```diff
@@ -0,0 +1,197 @@
...
+   * @param array|null $expected_image
+   *   The expected image values.
+   * @param string $variant
+   *   The variant of the pattern being checked.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertCardImage($expected_image, string $variant, Crawler $crawler): void {
+    if ($variant === 'search') {
+      $image_div = $crawler->filter('.row .col-md-3 img.card-img-top');
```
brummbar: Since the assertions are the same, can we just determine the selector in the if() ?

tests/src/PatternAssertion/CardPatternAssert.php
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * Asserts the card date block.
+   *
+   * @param array $expected
+   *   The expected data.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertDate(array $expected, Crawler $crawler): void {
+    // @todo Use dedicated pattern assert once we re-work date_plock pattern.
```
brummbar: date_block

tests/src/PatternAssertion/CardPatternAssert.php
brummbar: But it's too risky like this unfortunately, we need to clean it up.

tests/src/PatternAssertion/CardPatternAssert.php
```diff
@@ -0,0 +1,197 @@
...
+  /**
+   * Returns the base CSS selector for a list item depending on the variant.
+   *
+   * @param string $variant
+   *   The variant being checked.
+   *
+   * @return string
+   *   The base selector for the variant.
+   */
+  protected function getBaseItemClass(string $variant): string {
```
brummbar: Do we really need this separate? It's only used once.

brummbar (changes requested):

templates/patterns/card/card.ui_patterns.yml
```diff
@@ -35,6 +35,10 @@ card:
       preview:
         src: 'https://placeimg.com/1000/500/people'
         alt: 'Alternative text for card image'
+    date:
+      type: array
+      label: 'Date object'
+      description: 'Date for card. Optional.'
```
brummbar: I didn't remember the ticket and it took me a while to remember how to use this. It's not really an object, it's an associative array, let's put some info in here. At least let's write that it uses the same parameters of the bcl date-block template.

brummbar (changes requested):
All good, just this documentation change.

brummbar (approved):

PR #296, opened 2023-01-27
Drupal 10 compatiblity

claudiu-cristea:


brummbar:
Will be fixed in #298

PR #298, opened 2023-02-15
EWPP-3013: Drupal 10 compatibility.

brummbar:


claudiu-cristea (approved):
Doesn't seem to break backwards compatibility, so the tagging `1.x` for a new release looks OK to me.

drishu (approved):

PR #304, opened 2023-02-26
OEL-2249: Pager aria-label accessibility issue fix.

### LINKED ISSUE #277: Accessibility issue with pager
The pager currently has accessibility issues, here is the report from AXE (used during behat on SRB):

```
│ -
│   description: 'Ensures links have discernible text'
│   help: 'Links must have discernible text'
│   helpUrl: 'https://dequeuniversity.com/rules/axe/4.4/link-name?application=axeAPI'
│   id: link-name
│   impact: serious
│   nodes:
│     -
│       all: {  }
│       any:
│         -
│           data: null
│           id: has-visible-text
│           impact: minor
│           message: 'Element does not have text that is visible to screen readers'
│           relatedNodes: {  }
│         -
│           data: null
│           id: aria-label
│           impact: serious
│           message: 'aria-label attribute does not exist or is empty'
│           relatedNodes: {  }
│         -
│           data: null
│           id: aria-labelledby
│           impact: serious
│           message: 'aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty'
│           relatedNodes: {  }
│         -
│           data:
│             messageKey: noAttr
│           id: non-empty-title
│           impact: serious
│           message: 'Element has no title attribute'
│           relatedNodes: {  }
│       failureSummary: |-
│         Fix all of the following:
│           Element is in tab order and does not have accessible text
│ 
│         Fix any of the following:
│           Element does not have text that is visible to screen readers
│           aria-label attribute does not exist or is empty
│           aria-labelledby attribute does not exist, references elements that do not exist or references elements that are empty
│           Element has no title attribute
│       html: '<a class="page-link" href="?page=4"><svg class="bi icon--fluid"><use xlink:href="/themes/contrib/oe_bootstrap_theme/assets/icons/bcl-default-icons.svg#chevron-double-right"></use></svg></a>'
│       impact: serious
│       none:
│         -
│           data: null
│           id: focusable-no-name
│           impact: serious
│           message: 'Element is in tab order and does not have accessible text'
│           relatedNodes: {  }
│       target:
│         - '.page-item:nth-child(7) > .page-link[href="?page=4"]'
│   tags:
│     - cat.name-role-value
│     - wcag2a
│     - wcag412
│     - wcag244
│     - section508
│     - section508.22.a
│     - ACT
│ 
Critical/serious HTML Accessibility error(s) found. (Exception)
```

BCL allows passing aria labels since v0.18.0, see [here](https://github.com/openeuropa/bootstrap-component-library/commit/fb27d027cc298afbd9d4410eca876169da1ca0ae).

Perhaps we can simply pass hardcoded aria labels in `templates/overrides/navigation/pager.html.twig`?

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/overrides/navigation/pager.html.twig
+++ b/templates/overrides/navigation/pager.html.twig
@@ -11,38 +11,23 @@
     'first': items.first ? {
       'label': items.first.text|default('First'|t),
       'path': items.first.href,
-      'attributes': items.first.attributes ?
-                    items.first.attributes.setAttribute('aria-label', items.first.aria_label|default('First'|t))
-                    : create_attribute().setAttribute('aria-label', items.first.aria_label|default('First'|t)),
     },
     'prev': items.previous ? {
       'label': items.previous.text|default('Previous'|t),
       'path': items.previous.href,
-      'attributes': items.previous.attributes ?
-                    items.previous.attributes.setAttribute('aria-label', items.previous.aria_label|default('Previous'|t))
-                    : create_attribute().setAttribute('aria-label', items.previous.aria_label|default('Previous'|t)),
     },
     'items': items.pages|map((item, key) => {
       'label': key,
       'path': item.href,
       'active': key == current,
-      'attributes': item.attributes ? 
-                    item.attributes.setAttribute('aria-label', item.aria_label|default(key))
-                    : create_attribute().setAttribute('aria-label', item.aria_label|default(key)),
     }),
     'next': items.next ? {
       'label': items.next.text|default('Next'|t),
       'path': items.next.href,
-      'attributes': items.next.attributes ?
-                    items.next.attributes.setAttribute('aria-label', items.next.aria_label|default('Next'|t))
-                    : create_attribute().setAttribute('aria-label', items.next.aria_label|default('Next'|t)),
     },
     'last': items.last ? {
       'label': items.last.text|default('Last'|t),
       'path': items.last.href,
-      'attributes': items.last.attributes ?
-                    items.last.attributes.setAttribute('aria-label', items.last.aria_label|default('Last'|t))
-                    : create_attribute().setAttribute('aria-label', items.last.aria_label|default('Last'|t)),
     },
   }) }}
 {% endif %}
--- a/templates/overrides/navigation/pager.html.twig
+++ b/templates/overrides/navigation/pager.html.twig
@@ -11,28 +11,28 @@
     'first': items.first ? {
       'label': items.first.text|default('First'|t),
       'path': items.first.href,
-      'attributes': items.first.attributes.setAttribute('aria-label', 'First'|t),
+      'attributes': items.first.attributes ? items.first.attributes.setAttribute('aria-label', 'First'|t) : create_attribute().setAttribute('aria-label', 'First'|t),
     },
     'prev': items.previous ? {
       'label': items.previous.text|default('Previous'|t),
       'path': items.previous.href,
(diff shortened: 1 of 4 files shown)


templates/overrides/navigation/pager.html.twig
```diff
@@ -11,23 +11,38 @@
     'first': items.first ? {
       'label': items.first.text|default('First'|t),
       'path': items.first.href,
+      'attributes': items.first.attributes ?
```
brummbar: Why these are defined here too? We have the defaults in the pattern and we are using the template, so here no changes should be added.

templates/patterns/pagination/pattern-pagination.html.twig
```diff
@@ -4,18 +4,45 @@
  * Pagination pattern.
  */
 #}
+
+{% set internal_items = [] %}
+
```
brummbar: Drop the empty line to keep them close.

templates/patterns/pagination/pattern-pagination.html.twig
```diff
@@ -4,18 +4,45 @@
...
+{% for item in items %}
+  {% set internal_items = internal_items|merge([item|merge({
+      'attributes': item.attributes ?
+                    item.attributes.setAttribute('aria-label', item.aria_label|default(item.label))
+                    : create_attribute().setAttribute('aria-label', item.aria_label|default(item.label)),
+    })])
```
brummbar: A more concise version is:
```suggestion
      'attributes': create_attribute({'aria-label': item.label}).merge(item.attributes|default(create_attribute()))
    })])
```
First we create a new attribute object with the default value, then we merge it to the existing one (so we don't override items that already exist). In case there is no existing one, we create it.

tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -94,6 +94,7 @@ protected function assertLinks(array $expected, Crawler $crawler): void {
       if (isset($expected_item['label'])) {
         $this->assertSame($expected_item['label'], $link->html());
       }
+      $this->assertNotEmpty($link->attr('aria-label'));
```
brummbar: Here you need to check if in the expected data there's an `aria_label` entry and if yes, use `assertEquals` with the attribute value.
Also edit the tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml test file.
In the first scenario, do not add any aria labels in the "render" part (which is what we pass to the test), but add all the labels in the assertions part (which is what you test above).
Add also another scenario `pagination_with_aria` where you put some existing aria labels around and in others not, so we test the overriding part.

brummbar (changes requested):

templates/overrides/navigation/pager.html.twig
tibi2303: Removed them.

templates/patterns/pagination/pattern-pagination.html.twig
tibi2303: Removed.

templates/patterns/pagination/pattern-pagination.html.twig
tibi2303: Indeed, thank you. I was thinking on how to make it smaller and your suggestion was apprecieted. I changed all them accordingly.

tests/src/PatternAssertion/PaginationPatternAssert.php
tibi2303: Done. I see the tests are passing.

brummbar (approved):

PR #307, opened 2023-03-02
OEL-2251: Use heading in featured-media pattern.

### LINKED ISSUE #288: Only display heading for title in the pattern featured media if it is not empty
The pattern Featured media always displays the title headings, even if the title is empty (see [here](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.0.0-beta7/templates/patterns/featured_media/pattern-featured-media.html.twig#L27))

It should only display the heading if the title is not empty. Like this:

```
{% if title %}
  <h2 class="mb-4">{{ title }}</h2>
{% endif %}
```

This is even more important since the title is not required by default (see [here](https://github.com/openeuropa/oe_paragraphs/blob/1.17.0/config/install/field.storage.paragraph.field_oe_title.yml)).

Also it would be nice to have the "title_tag" parameter available.

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/featured_media/featured_media.ui_patterns.yml
+++ b/templates/patterns/featured_media/featured_media.ui_patterns.yml
@@ -35,6 +35,14 @@ featured_media:
       label: "Heading"
       description: "The heading, displayed on the top."
       preview: "Lorem ipsum"
+    title_tag:
+      type: text
+      label: 'Title tag'
+      description: 'The tag to use for the title. Defaults to h2.'
+    title_link:
+      type: LinkObject
+      label: 'Title Link'
+      description: 'Link Object including path to set href. Label is set by title attribute.'
     subtitle:
       type: text
       label: 'Subtitle'
--- a/templates/patterns/featured_media/featured_media.ui_patterns.yml
+++ b/templates/patterns/featured_media/featured_media.ui_patterns.yml
@@ -39,10 +39,10 @@ featured_media:
       type: text
       label: 'Title tag'
       description: 'The tag to use for the title. Defaults to h2.'
-    title_link:
-      type: LinkObject
-      label: 'Title Link'
-      description: 'Link Object including path to set href. Label is set by title attribute.'
+    title_url:
+      type: text
+      label: 'Title URL'
+      description: 'The link of the title.'
     subtitle:
       type: text
       label: 'Subtitle'
--- a/templates/patterns/featured_media/featured_media.ui_patterns.yml
+++ b/templates/patterns/featured_media/featured_media.ui_patterns.yml
@@ -35,14 +35,6 @@ featured_media:
       label: "Heading"
       description: "The heading, displayed on the top."
       preview: "Lorem ipsum"
-    title_tag:
-      type: text
-      label: 'Title tag'
-      description: 'The tag to use for the title. Defaults to h2.'
-    title_url:
-      type: text
-      label: 'Title URL'
-      description: 'The link of the title.'
     subtitle:
       type: text
       label: 'Subtitle'
--- a/templates/patterns/featured_media/pattern-featured-media.html.twig
+++ b/templates/patterns/featured_media/pattern-featured-media.html.twig
@@ -24,7 +24,14 @@
       {% set video_alignment_class = 'order-md-1' %}
     {% endif %}
 
-    <h2 class="mb-4">{{ title }}</h2>
+    {%- if title is not empty -%}
+      {% include '@oe-bcl/heading' with {
+        title: title,
+        title_tag: title_tag|default('h2'),
+        title_link: title_link|default(''),
+        attributes: create_attribute({'class': 'mb-4'}).merge(item.attributes|default(create_attribute())),
+      } only %}
+    {%- endif -%}
     <div class="row">
       <div class="col-12 col-md-6 {{ text_alignment_class }}">
         {% if subtitle %}
--- a/templates/patterns/featured_media/pattern-featured-media.html.twig
+++ b/templates/patterns/featured_media/pattern-featured-media.html.twig
@@ -28,8 +28,10 @@
       {% include '@oe-bcl/heading' with {
         title: title,
         title_tag: title_tag|default('h2'),
-        title_link: title_link,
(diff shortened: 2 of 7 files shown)


templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -24,7 +24,14 @@
       {% set video_alignment_class = 'order-md-1' %}
     {% endif %}
 
-    <h2 class="mb-4">{{ title }}</h2>
+    {%- if title is not empty -%}
+      {% include '@oe-bcl/heading' with {
+        title: title,
+        title_tag: title_tag|default('h2'),
```
drishu: the type of the `title` field in `featured_media.ui_patterns.yml` is text, there is no title_tag nor title_link

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -24,7 +24,14 @@
...
+        title_link: title_link|default(''),
```
drishu: here we have the choice between an undefined variable and an empty string, does it make sense to set it?

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -24,7 +24,14 @@
...
+        attributes: create_attribute({'class': 'mb-4'}).merge(item.attributes|default(create_attribute())),
```
drishu: calling attributes on an undefined variable is redundant

drishu (changes requested):

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: indeed but it should have been title.attributes, not item.attributes.. changed it.

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: I removed the default on title_link.

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: I added also title_tag and title_link in yml file.

templates/patterns/featured_media/featured_media.ui_patterns.yml
```diff
@@ -35,6 +35,14 @@ featured_media:
       label: "Heading"
       description: "The heading, displayed on the top."
       preview: "Lorem ipsum"
+    title_tag:
```
drishu: we have a precedent in gallery and since its BC I dont see why not.

templates/patterns/featured_media/featured_media.ui_patterns.yml
```diff
@@ -35,6 +35,14 @@ featured_media:
...
+      type: text
+      label: 'Title tag'
+      description: 'The tag to use for the title. Defaults to h2.'
+    title_link:
+      type: LinkObject
```
drishu: The type must be a php type, like string, array, or an object of a certain class name, this is neither. Let's use string and set whatever format the heading in the template. Then also name it correspondingly, perhaps title_url ?

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -24,7 +24,14 @@
       {% set video_alignment_class = 'order-md-1' %}
     {% endif %}
 
-    <h2 class="mb-4">{{ title }}</h2>
+    {%- if title is not empty -%}
+      {% include '@oe-bcl/heading' with {
+        title: title,
+        title_tag: title_tag|default('h2'),
+        title_link: title_link,
```
drishu: with my previous comment I meant that the entire line can dissapear, but since you added it to the yml, and I dont see why not, then you can add the default back

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -24,7 +24,14 @@
...
+        attributes: create_attribute({'class': 'mb-4'}).merge(title.attributes|default(create_attribute())),
```
drishu: title.attributes is still nok, because title is marked as a string in the yml, and we should keep it that way.. we want to move away from these objects because they cause complications in the drupal layer that limit our solutions.

drishu (changes requested):

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: it should be cleaned now. Only creating the attribute and adding the class.

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: It should not only be the default as this param is expecting a BCL link object. I added the check if it's exists and merging it accordingly. Also did this for gallery.

templates/patterns/featured_media/featured_media.ui_patterns.yml
tibi2303: Indeed. I also added tests and changed the pattern of gallery. Please have a look.  I guess we either change the ticket and title of this PR or we make the gallery changes in another ticket.

templates/patterns/featured_media/featured_media.ui_patterns.yml
tibi2303: Yes. changed

templates/patterns/gallery/pattern-gallery.html.twig
```diff
@@ -12,6 +12,12 @@
 {% set items = bcl_gallery_items(items|default([])) %}
 {% set icon_path = bcl_icon_path %}
 
+{%- if title_url is not empty -%}
```
drishu: I think its a misunderstanding, let's rather remove all these new fields to prevent opening a pandoras box of heading fields. And later if they are needed it will be a separate ticket handing them all

drishu (changes requested):

templates/patterns/gallery/pattern-gallery.html.twig
tibi2303: done.

drishu (approved):

PR #309, opened 2023-03-06
OEL-2253: Adapt content_banner template for badges with outline param.

### LINKED ISSUE #293: Make it possible to set the outline badge parameter in content banner
The pattern `content_banner` displays the badges with the `outline` property set to `true` by default and it cannot be overridden: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/content_banner/pattern-content-banner.html.twig#L22

We can keep this behavior but if a badge should be displayed with `'outline': false,` in the content banner then something like this would be better:
```
{# Enable 'outline' in every badge if it is not explicitly set. #}
{% set _badges = [] %}
{% for _badge in badges %}
  {% if _badge.outline is not defined %}
    {% set _badge = _badge|merge({
      'outline': true,
    }) %}
  {% endif %}
  {% set _badges = _badges|merge([_badge]) %}
{% endfor %}
```

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -23,7 +23,7 @@
 {% set _badges = [] %}
 {% for _badge in badges %}
   {% set _badge = _badge|merge({
-    'outline': _badge.outline ?? true,,
+    'outline': _badge.outline ?? true,
   }) %}
   {% set _badges = _badges|merge([_badge]) %}
 {% endfor %}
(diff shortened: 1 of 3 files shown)


templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -23,7 +23,7 @@
 {% set _badges = [] %}
 {% for _badge in badges %}
   {% set _badge = _badge|merge({
-    'outline': true,
+    'outline': _badge.outline|default(true),
```
brummbar: This doesn't work, it always shows outlined entries:
image
The correct check is `'outline': _badge.outline ?? true,` to preserve false values.

brummbar (changes requested):

templates/patterns/content_banner/pattern-content-banner.html.twig
tibi2303: yes, thanks. changed them

brummbar (approved):

PR #310, opened 2023-03-06
OEL-2254: Fixed path of bootstrap-ie11.css

### LINKED ISSUE #202: The bootstrap-ie11.css is included from the active theme
In `includes/page_attachements.inc` the active theme path is used to include the bootstrap-ie11.css file. Is there a reason for this? We need to copy this file because of this to our theme's assets folder, although it could be included from OE Bootstrap theme's assets folder as well.

See: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.0.0-alpha7/includes/page_attachments.inc#L19-L22

tibi2303:
Jira issue(s):

### DIFF
--- a/includes/page_attachments.inc
+++ b/includes/page_attachments.inc
@@ -16,8 +16,8 @@ use Drupal\Core\Render\Markup;
  */
 function oe_bootstrap_theme_page_attachments_alter(array &$attachments) {
   global $base_url;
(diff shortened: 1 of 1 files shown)


drishu (dismissed):
We don't need access checks and we know the file is present, so its ok for me. I also don't think this needs a test.

drishu (approved):

drishu (approved):

PR #315, opened 2023-03-15
OEL-2261: Override unformatted pattern for correct layout margins.

### LINKED ISSUE #214: Card margin bottom
I wonder why the card pattern is hardcoding the margin to the bottom?
https://github.com/openeuropa/oe_bootstrap_theme/blob/1.0.0-alpha7/templates/patterns/card/pattern-card--variant-search.html.twig#L56

I think the pattern should set the layout for the card itself only. If spacing is needed between a list of cards, we should leave this to the list itself? This is how it is implemented in BCL, see here https://oelibrary.netlify.app/ucpkn/?path=/story/features-project--listing

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/overrides/views/views-view-unformatted.html.twig
+++ b/templates/overrides/views/views-view-unformatted.html.twig
@@ -1,0 +1,22 @@
+{#
+/**
+ * @file
+ * Template for a unformatted rows.
+ */
+#}
+{% if title %}
+  <h3>{{ title }}</h3>
+{% endif %}
+<div class="row row-cols-1 g-4">
+{% for row in rows %}
+  {%
+    set row_classes = [
+      'col',
+      default_row_class ? 'views-row',
+    ]
+  %}
+  <div{{ row.attributes.addClass(row_classes) }}>
+    {{- row.content -}}
+  </div>
+{% endfor %}
+</div>
--- a/templates/patterns/card/pattern-card--variant-search.html.twig
+++ b/templates/patterns/card/pattern-card--variant-search.html.twig
@@ -49,7 +49,6 @@
     'border-bottom',
     'border-md-0',
     'border-0',
-    'mb-3',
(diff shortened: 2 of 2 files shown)


templates/overrides/views/views-view-unformatted.html.twig
```diff
@@ -0,0 +1,22 @@
+{#
+/**
+ * @file
+ * Template for a unformatted rows.
```
drishu, line 4: The english here is not correct, but I would change it completely to "Theme override to display a view of unformatted rows."

templates/overrides/views/views-view-unformatted.html.twig
```diff
@@ -0,0 +1,22 @@
...
+ */
+#}
+{% if title %}
```
drishu, line 7: would the bcl heading template be suited here?

templates/overrides/views/views-view-unformatted.html.twig
```diff
@@ -0,0 +1,22 @@
...
+  <h3>{{ title }}</h3>
+{% endif %}
+<div class="row row-cols-1 g-4">
+{% for row in rows %}
+  {%
+    set row_classes = [
+      'col',
```
drishu, line 14: the result of this is not the same as mb-3, I checked, the spacing is not the same

drishu (commented):
I don't think its mandatory but I agree with the principle that the list can define the margins.
That being said we have to be sure its backwards compatible, and there could be other types of views (other than unformatted) where this variant has been used. Also this template might already been overridden.
This looks rather like a ticket for V2, or maybe (not sure its worth it) we can add a parameter that removes the mb-3.

brummbar:
I agree with @drishu , we cannot remove the class without changing the behaviour for existing websites.

PR #318, opened 2023-03-19
OEL-2262: Improvements on content-banner pattern.

### LINKED ISSUE #314: Content banner pattern fixes and improvements
Suggested improvement for the content banner pattern:
1. Make it possible to set the title tag.
2. Do not display span wrapper when a [_meta_item](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/content_banner/pattern-content-banner.html.twig#L15) is empty .
3. Do not display div wrapper when the [content ](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/content_banner/pattern-content-banner.html.twig#L19)is empty.
4. Add new action_bar field and pass it to the BCL component.

We are currently overriding the pattern because of the issues above. See here https://github.com/digit-devs/echo-ucpkn-dev/blob/master/lib/themes/ucpkn_theme/templates/overrides/patterns/content_banner/pattern-content-banner.html.twig

tibi2303:
Jira issue(s):

### DIFF
--- a/tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
+++ b/tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
@@ -1,4 +1,4 @@
-content_banner_white_background_with_image_with_badges_with_meta_with_links:
+content_banner_white_background_with_image_with_badges_with_meta_with_links_with_action_bar_with_title_tag:
   render:
     '#type': pattern
     '#id': 'content_banner'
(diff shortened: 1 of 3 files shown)


drishu (changes requested):
looks good but we should cover this in the tests

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
```diff
@@ -28,8 +31,9 @@ content_banner_white_background_with_image_with_badges_with_meta_with_links:
       'div.bcl-content-banner': 1
       'div.bcl-content-banner.bg-white': 1
       'img': 1
+      'div.bcl-content-banner > div.container > div.action-bar': 1
```
drishu, line 34: we should also check for the button you added

drishu (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/content_banner.yml
tibi2303: Changed it to anchor tag. I don't understand why the phpunit is not rendering the button. It should not be a problem with the template or the new parameter.

drishu (approved):

PR #322, opened 2023-03-22
OEL-2266: Added image on mobile and fixed spacing in card search.

### LINKED ISSUE #319: The styles for the card pattern (search variant) and BCL are not consistent
It seems that the styles applied to the card pattern (search variant) is not aligned with the styles in the BCL.
### **Card pattern:**
```
<div class="row">
<div class="col-md-3 col-lg-2 rounded mw-listing-img">
<div class="col-md-9 col-lg-10">
</div>
```
**Font style: fs-4**
**_No space between the image/date block and the title_**
https://digit-oesa.acc.fpfis.tech.ec.europa.eu/en/node/47
image

------------------------------------------------------------------------
### **BCL:**
```
<div class="row">
<div class="bcl-card-start-col mb-3 mb-lg-0 bcl-size-small">
<div class="col-12 col-md">
</div>
```
**Font style: fs-5**
image

**_I think the card pattern should be updated to align with BCL_**

### LINKED ISSUE #320: Card pattern: Support displaying image on small screen
For some search pages (users page, group member), the client would like to display the image on small screen. This is not possible with the current search variant card pattern due to [this](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/card/pattern-card--variant-search.html.twig#L57) We have fixed it in ucpkn by adding image_display for the card pattern. See [here](https://github.com/digit-devs/echo-ucpkn-dev/pull/1133)

It would be nice to support the image display on the different breakpoints.

tibi2303:
Jira issue(s):

### DIFF
--- a/config/schema/oe_bootstrap_theme.schema.yml
+++ b/config/schema/oe_bootstrap_theme.schema.yml
@@ -12,3 +12,10 @@ oe_bootstrap_theme.settings:
         responsive:
           type: string
           label: 'Responsive tables'
+    oel_improvements:
+      type: mapping
+      label: 'OEL Improvements'
+      mapping:
+        oel_2266:
+          type: boolean
+          label: 'OEL 2266'
--- a/config/schema/oe_bootstrap_theme.schema.yml
+++ b/config/schema/oe_bootstrap_theme.schema.yml
@@ -16,7 +16,7 @@ oe_bootstrap_theme.settings:
       type: mapping
       label: 'OEL Improvements'
       mapping:
-        card_search_image_not_visible_on_mobile:
+        card_search_image_hide_on_mobile:
           type: boolean
           label: 'Card image not visible on mobile'
         card_search_use_grid_classes:
--- a/config/schema/oe_bootstrap_theme.schema.yml
+++ b/config/schema/oe_bootstrap_theme.schema.yml
@@ -12,7 +12,7 @@ oe_bootstrap_theme.settings:
         responsive:
           type: string
           label: 'Responsive tables'
-    oebt_bc:
+    backward_compatibility:
       type: mapping
       label: 'OEL Improvements'
       mapping:
--- a/config/schema/oe_bootstrap_theme.schema.yml
+++ b/config/schema/oe_bootstrap_theme.schema.yml
@@ -12,10 +12,13 @@ oe_bootstrap_theme.settings:
         responsive:
           type: string
           label: 'Responsive tables'
-    oel_improvements:
+    oebt_bc:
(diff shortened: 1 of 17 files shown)


templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -54,7 +54,6 @@
   'text_color': 'dark',
   'image': (image is not empty) ? image|merge({
     'position': 'top',
-    'classes': 'd-none d-md-block',
```
drishu, line 59: I never understood why the image needs to disappear in mobile view, but now that we change, is this a bug fix or a change request?

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -33,8 +33,8 @@
 {% include '@oe-bcl/card' with {
   'horizontal': true,
   'horizontal_grid': {
-    'left_col_classes': 'col-md-3 col-lg-2 rounded mw-listing-img',
```
drishu, line 38: unfortunately this now changes the size of the left column in the card, and with it the image and date widget, we need to discuss in the team if this is a breaking change or not.

drishu (commented):
these 2 points I need to discuss

includes/pattern.inc
```diff
@@ -284,3 +284,11 @@ function oe_bootstrap_theme_preprocess_pattern_gallery__preview(array &$variable
     }
   }
 }
+
+/**
+ * Implements hook_preprocess_HOOK() for gallery pattern.
+ */
+function oe_bootstrap_theme_preprocess_pattern_card(array &$variables): void {
+  $variables['oebt_bc_card_search_image_not_visible_on_mobile'] = theme_get_setting('oebt_bc.card_search_image_not_visible_on_mobile') ?? TRUE;
```
brummbar: What about?
```suggestion
  $variables['oebt_bc_card_search_image_hidden_on_mobile'] = theme_get_setting('oebt_bc.card_search_image_hidden_on_mobile') ?? TRUE;
```

config/schema/oe_bootstrap_theme.schema.yml
```diff
@@ -12,3 +12,13 @@ oe_bootstrap_theme.settings:
         responsive:
           type: string
           label: 'Responsive tables'
+    backward_compatibility:
+      type: mapping
+      label: 'OEL Improvements'
```
drishu, line 17: Not important but I forgot to update this label

drishu (approved):

PR #331, opened 2023-05-24
[pattern_file] Open PDF file to other tab.

julien-:
# What was not OK
PDF Files are opened in the current tab.

# What is done here
When PDF file : Add attribute `target="_blank"` and Remove `download` attribute.

All other changes are just alphabetical ordering functions and moving preprocess_page to its dedicated file.

### DIFF
--- a/includes/page.inc
+++ b/includes/page.inc
@@ -1,0 +1,17 @@
+<?php
+
+/**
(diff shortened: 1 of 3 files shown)


julien-:
Escalated to BCL

PR #337, opened 2023-07-07
ELA-667: Create patch for oe_bootstrap_theme_helper... TwigExtension.…

libetho:
…php => bclCardList().

Jira issue(s):


brummbar:
The work added in this PR is dangerous and can cause XSS attacks.

PR #344, opened 2023-07-24
UCPKN-2311: Replace placeimg with picsum in previews and tests.

kp77:
Jira issue(s):
- [UCPKN-2311](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/UCPKN-2311)


tibi2303:
Changed to picsum as it works and supports more params than loremflickr and also it's the same as on bcl, we've updated for gallery as on loremflickr and it's ok now.

brummbar (approved):

PR #348, opened 2023-08-02
ELA-751: Contextual classes.

julien-:


brummbar (changes requested):
I need more context in order to understand what a PR does.
Also, it misses tests.

PR #350, opened 2023-08-07
ELA-665: OE Featured Media formatter with Image style.

julien-:
### Change log

- Added:
1. Possibility tu use the Drupal rendering using **oe_featured_media** field as an Image with image_style
2. You can use "UI Patterns Layouts" module to rennder the **oe_featured_media** field as an ImageValueObject  

Requires https://github.com/openeuropa/oe_content/pull/593


brummbar (changes requested):
Formatters belong to oe_whitelabel rather than oe_bootstrap_theme.
In any case, we are planning to rework completely some patterns, including the media one.
PR misses tests.

julien-:
See https://github.com/openeuropa/oe_whitelabel/pull/254/files

PR #361, opened 2023-11-22
OEL-2742: Gallery captions hide on video.

### LINKED ISSUE #327: Gallery pattern: caption title, link is rendered for playable media
In the [BCL example](https://oelibrary.netlify.app/default/?path=/story/paragraphs-gallery--default), the gallery component never has a caption overlay for videos. This is probably because with the caption overlay, the video controls are not usable.

In the Gallery pattern, this is [handled](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.3.0/templates/patterns/gallery/pattern-gallery.html.twig#L25), but only for the `item.caption` property. If the item has `caption_title` or `link` property, the caption overlay is displayed, making the video controls unusable.

The issue is mitigated by the fact that we can map the video media name to the `caption` property instead of the `caption_title`, but I think the correct solution would be if the caption overlay is never displayed for playable media in the pattern.

Suggestion for https://github.com/openeuropa/oe_bootstrap_theme/blob/1.3.0/templates/patterns/gallery/pattern-gallery.html.twig#L24-L26:

```
    items: items|map(_item => _item|merge({
      caption: _item.is_playable ? '' : _item.caption,
      caption_title: _item.is_playable ? '' : _item.caption_title,
      link: _item.is_playable ? '' : _item.link
    })),
```
Or another option would be to move this logic to the [bcl_gallery_items](https://github.com/openeuropa/oe_bootstrap_theme/blob/1.3.0/modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php#L299) twig function.

tibi2303:
Jira issue(s):

### DIFF
--- a/tests/src/PatternAssertion/CarouselPatternAssert.php
+++ b/tests/src/PatternAssertion/CarouselPatternAssert.php
@@ -104,28 +104,25 @@ class CarouselPatternAssert extends BasePatternAssert {
       try {
         self::assertStringContainsString($expected_item['image'], $item->html());
         $this->assertElementText($expected_item['caption_title'] ?? NULL, '.carousel-caption .fs-5', $item);
+        $this->assertElementAttribute($expected_item['interval'] ?? 0, '.carousel-item', 'data-bs-interval', $item);
         if ($isPlayable) {
           $this->assertElementNotExists('.carousel-caption', $item);
+          continue;
+        }
+        if (isset($expected_item['caption'])) {
+          $this->assertElementTextContains($expected_item['caption'], '.carousel-caption', $item);
+        }
+        if (isset($expected_item['caption_classes'])) {
+          $this->assertElementExists('.carousel-caption > .' . $expected_item['caption_classes'], $item);
+        }
+        if (isset($expected_item['link'])) {
+          $this->assertElementText($expected_item['link']['label'], '.carousel-caption a', $item);
+          $this->assertElementAttribute($expected_item['link']['path'], '.carousel-caption a', 'href', $item);
+          $this->assertElementExists('.carousel-caption a svg', $item);
         }
         else {
-          if (isset($expected_item['caption'])) {
-            $this->assertElementTextContains($expected_item['caption'], '.carousel-caption', $item);
-          }
-          if (isset($expected_item['caption_classes'])) {
-            $this->assertElementExists('.carousel-caption > .' . $expected_item['caption_classes'], $item);
-          }
-          if (isset($expected_item['link'])) {
-            $this->assertElementText($expected_item['link']['label'], '.carousel-caption a', $item);
-            $this->assertElementAttribute($expected_item['link']['path'], '.carousel-caption a', 'href', $item);
-            $this->assertElementExists('.carousel-caption a svg', $item);
-          }
-          else {
-            $this->assertElementNotExists('.carousel-caption a', $item);
-          }
+          $this->assertElementNotExists('.carousel-caption a', $item);
         }
-
-        $this->assertElementAttribute($expected_item['interval'] ?? 0, '.carousel-item', 'data-bs-interval', $item);
-
       }
       catch (\Exception $e) {
         throw new \Exception(sprintf('Failed asserting data for item %s.', $index), 0, $e);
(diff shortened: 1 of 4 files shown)


tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -98,28 +98,34 @@ protected function assertItems(array $expected, Crawler $crawler): void {
...
+      // Check if HTML content contains video or iframe.
+      $isPlayable = strpos($item->html(), '<video') !== FALSE || strpos($item->html(), '<iframe') !== FALSE;
+
       try {
         self::assertStringContainsString($expected_item['image'], $item->html());
         $this->assertElementText($expected_item['caption_title'] ?? NULL, '.carousel-caption .fs-5', $item);
-
-        if (isset($expected_item['caption'])) {
-          $this->assertElementTextContains($expected_item['caption'], '.carousel-caption', $item);
+        if ($isPlayable) {
```
AaronGilMartinez, line 108: I think we can improve this conditionals a little bit. What I would do is:
- Move last assert ( line 127) to the beginning, before all the ifs so it's always checked.
- Replace the `if ($isPlayable) {` else with a `continue;` inside the if.

AaronGilMartinez (changes requested):
Small change

AaronGilMartinez (dismissed):

tests/src/PatternAssertion/GalleryPatternAssert.php
```diff
@@ -83,6 +83,7 @@ protected function assertItems(array $expected, Crawler $crawler): void {
         'image' => $item['media']['rendered'],
         'caption_title' => $item['media']['caption_title'] ?? NULL,
         'caption' => $item['media']['caption_media'] ?? NULL,
+        'is_playable' => $item['media']['is_playable'] ?? NULL,
```
drishu: Seems you wanted to use this, I think your idea was to set this send to the carousel expected items and asset there that it matches? anyway, in the end you did not set any actual values to it, nor are you using it in the carousel pattern.

tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -98,20 +98,23 @@ protected function assertItems(array $expected, Crawler $crawler): void {
     foreach ($expected as $index => $expected_item) {
       $item = $items->eq($index);
 
+      // Check if HTML content contains video or iframe.
```
drishu:
```suggestion
      // Check if the HTML content contains video or iframe.
```

tests/src/PatternAssertion/CarouselPatternAssert.php
```diff
@@ -98,20 +98,23 @@ protected function assertItems(array $expected, Crawler $crawler): void {
...
+      $isPlayable = strpos($item->html(), '<video') !== FALSE || strpos($item->html(), '<iframe') !== FALSE;
+
       try {
         self::assertStringContainsString($expected_item['image'], $item->html());
         $this->assertElementText($expected_item['caption_title'] ?? NULL, '.carousel-caption .fs-5', $item);
-
+        $this->assertElementAttribute($expected_item['interval'] ?? 0, '.carousel-item', 'data-bs-interval', $item);
```
drishu, line 107: an $expected_item['interval'] is never set

drishu (changes requested):

tests/src/PatternAssertion/CarouselPatternAssert.php
tibi2303: i made a dump for this on test 'carousel_all_item_values' and it's populated on my side.
I made another dump for one which is not set and it's undefined. So it works as expected.

tests/src/PatternAssertion/GalleryPatternAssert.php
tibi2303: Indeed, i removed it because it's already tested inside the carousel test directly using $isPlayable.

tests/src/PatternAssertion/CarouselPatternAssert.php
drishu: Its just that it was used in the carousel test, but not gallery, so it threw me off a bit, but its ok.

drishu (approved):

AaronGilMartinez (approved):

PR #362, opened 2023-12-12
OEL-2741: Fixed bcl_card_list twig filter.

tibi2303:
Jira issue(s):

### DIFF
--- a/modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
+++ b/modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
@@ -24,6 +24,9 @@ use Twig\TwigFunction;
 
 /**
  * Collection of extra Twig extensions as filters and functions.
+ *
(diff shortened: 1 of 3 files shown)


brummbar (changes requested):
I am pretty confidenti this approach is wrong.
It's not oe_bootstrap_theme that needs to wrap into Markup objects the values. It's the caller task.
For example, normally you pass a render array or an already escaped string. 
But with this approach, you trust anything that is being sent. If a dangerous string that contains a script tag is passed, it would be rendered straight away.

brummbar:
This is the output of the pattern when used in the list item block paragraph (oe_whitelabel) with the 1.x version of this theme:
image

brummbar:
This is what happens when I apply the current patch:

https://github.com/openeuropa/oe_bootstrap_theme/assets/8488617/7163dbf8-ff19-43bb-8283-e0acfa1bac57

Never use Markup::create(). Never pass html as string. Either pass render arrays, or the output of a render call.

PR #370, opened 2024-01-03
OEL-2783: Footer at the bottom of the screen.

julien-:


julien-:
Finally, have to override in final projects due to .dialog-off-canvas-main-canvas

PR #374, opened 2024-01-10
OEL-2787: Fix url options loss in bclCardList().

drishu:
Jira issue(s):

### DIFF
--- a/modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
+++ b/modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
@@ -459,7 +459,7 @@ TWIG;
   }
 
   /**
-   * Provides dara for testBclCardList().
+   * Provides data for testBclCardList().
    *
    * @return array
    *   The scenarios.
--- a/modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
+++ b/modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
@@ -9,6 +9,7 @@ use Drupal\Core\Render\BubbleableMetadata;
 use Drupal\Core\Render\RenderContext;
 use Drupal\Core\Template\Attribute;
 use Drupal\Core\Url;
+use Drupal\oe_bootstrap_theme\ValueObject\ImageValueObject;
 use Drupal\Tests\oe_bootstrap_theme\Kernel\AbstractKernelTestBase;
 use PHPUnit\Framework\ExpectationFailedException;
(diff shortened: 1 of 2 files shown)


modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
```diff
@@ -381,4 +382,143 @@ protected function elementChildrenFilterDataProvider(): array {
     return $scenarios;
   }
 
+  /**
+   * Tests BCL card list.
```
AaronGilMartinez: It could be seen like we are going to test card list, I would use the same description than other filter tests.

AaronGilMartinez (changes requested):
Little remark, the branch needs to be updated

modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
```diff
@@ -381,4 +382,143 @@ protected function elementChildrenFilterDataProvider(): array {
...
+
+      $this->assertEquals($expected['subtitle'], $result[$key]['subtitle'] ?? []);
+      $this->assertEquals($expected['text'], $result[$key]['text'] ?? []);
+      $this->assertEquals($expected['image'], $result[$key]['image'] ?? []);
+      $this->assertEquals($expected['badges'], $result[$key]['badges'] ?? []);
+    }
+  }
+
+  /**
+   * Provides dara for testBclCardList().
```
intelektron: Typo in `data`

intelektron (dismissed):

AaronGilMartinez (approved):

modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php
```diff
@@ -427,4 +428,143 @@ protected function elementChildrenFilterDataProvider(): array {
     return $scenarios;
   }
 
+  /**
+   * Tests BCL card list filter.
+   */
+  public function testBclCardList(): void {
+    $extension = $this->container->get('oe_bootstrap_theme_helper.twig_extension');
+    $data = $this->bclCardListDataProvider();
```
intelektron, line 436: Is there a reason to avoid the `@dataProvider` annotation?

brummbar: We tend to avoid data providers for tests from kernel and up, because it takes a huge time just to make a new install and the tested data is always so small.

intelektron (approved):

PR #379, opened 2024-02-01
OEL-2595: Fixed overlap of toolbar.

tibi2303:
Jira issue(s):

### DIFF
--- a/resources/sass/overrides/_toolbar.scss
+++ b/resources/sass/overrides/_toolbar.scss
@@ -4,6 +4,22 @@
 // Lower admin toolbar z-index in order to not overlap offcanvas and modal.
 #toolbar-administration {
   #toolbar-bar {
-    z-index: 61;
+    z-index: 5000;
   }
+}
+
+// Offcanvas
+.offcanvas-backdrop {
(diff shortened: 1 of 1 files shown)


resources/sass/overrides/_toolbar.scss
```diff
@@ -1,9 +1,9 @@
 // Drupal overrides for the admin toolbar
 // -----------------------------------------------------------------------------
 
-// Increase admin toolbar z-index
+// Lower admin toolbar z-index in order to not overlap offcanvas and modal.
 #toolbar-administration {
   #toolbar-bar {
-    z-index: 5000;
+    z-index: 61;
```
brummbar: Can't we do the opposite, and raise the offcanvas/modal?

brummbar (changes requested):

resources/sass/overrides/_toolbar.scss
tibi2303: done, francesco. I was thinking to not touch bs z-index but I've tested it and it seems these are the only places where they would overlap.

brummbar (approved):

PR #382, opened 2024-02-15
OEL-2794: Update components to run on Drupal 10.2.x.

AaronGilMartinez:
Jira issue(s):

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -19,7 +19,7 @@
         "drupal/core-composer-scaffold": "^10",
         "drupal/core-dev": "^10",
         "drupal/styleguide": "^2",
-        "drush/drush": "^11.1",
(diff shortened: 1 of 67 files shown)


composer.json
```diff
@@ -19,7 +19,7 @@
         "drupal/core-composer-scaffold": "^10",
         "drupal/core-dev": "^10",
         "drupal/styleguide": "^2",
-        "drush/drush": "^11.1",
```
brummbar, line 22: Let's take the chance to remove cweagans/composer-patches in line 10, we don't have patches declared.

brummbar (changes requested):

brummbar (approved):

PR #395, opened 2024-05-13
OEL-1771: Moved patterns tests from oe_whitelabel.

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -7,9 +7,9 @@
 {% apply spaceless %}
 {# Render the 'meta' items into the content variable. #}
 {% set _content %}
-  {% set content_attributes = create_attribute().addClass(['mt-4', 'card-text']) %}
+  {% set content_attributes = create_attribute().addClass(['mt-4', 'card-content']) %}
   {% if meta is not empty %}
-    {% set content_attributes = create_attribute().addClass('card-text') %}
+    {% set content_attributes = create_attribute().addClass('card-content') %}
     <div class="my-4">
       {% for _meta_item in meta %}
         {% if _meta_item is not empty %}
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -41,7 +41,7 @@
     'path': image.src,
     'alt': image.alt,
   } : {},
-  'image_size': image_size ?? 'lg',
+  'image_size': image_size|default('lg'),
   'badges': _badges,
   'links': links,
   'action_bar': action_bar,
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -41,7 +41,7 @@
     'path': image.src,
     'alt': image.alt,
   } : {},
-  'image_size': image_size|default('lg'),
+  'image_size': image_size ?? 'lg',
   'badges': _badges,
   'links': links,
   'action_bar': action_bar,
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -7,9 +7,9 @@
 {% apply spaceless %}
 {# Render the 'meta' items into the content variable. #}
 {% set _content %}
-  {% set content_attributes = create_attribute().addClass(['mt-4', 'banner-content']) %}
+  {% set content_attributes = create_attribute().addClass(['mt-4', 'card-text']) %}
   {% if meta is not empty %}
-    {% set content_attributes = create_attribute().addClass('banner-content') %}
+    {% set content_attributes = create_attribute().addClass('card-text') %}
     <div class="my-4">
       {% for _meta_item in meta %}
         {% if _meta_item is not empty %}
--- a/templates/patterns/content_banner/pattern-content-banner.html.twig
+++ b/templates/patterns/content_banner/pattern-content-banner.html.twig
@@ -7,9 +7,9 @@
 {% apply spaceless %}
 {# Render the 'meta' items into the content variable. #}
 {% set _content %}
-  {% set content_attributes = create_attribute().addClass('mt-4') %}
+  {% set content_attributes = create_attribute().addClass(['mt-4', 'banner-content']) %}
   {% if meta is not empty %}
-    {% set content_attributes = create_attribute() %}
+    {% set content_attributes = create_attribute().addClass('banner-content') %}
     <div class="my-4">
       {% for _meta_item in meta %}
         {% if _meta_item is not empty %}
--- a/tests/src/PatternAssertion/ContentBannerAssert.php
+++ b/tests/src/PatternAssertion/ContentBannerAssert.php
@@ -2,7 +2,7 @@
 
 declare(strict_types=1);
 
-namespace Drupal\Tests\oe_bootstrap_theme\PatternAssertions;
+namespace Drupal\Tests\oe_bootstrap_theme\PatternAssertion;
 
 use Symfony\Component\DomCrawler\Crawler;
 
@@ -20,18 +20,31 @@ class ContentBannerAssert extends BasePatternAssert {
         [$this, 'assertImage'],
         '.card-img-top',
       ],
+      'title' => [
+        [$this, 'assertElementText'],
+        '.card-title.bcl-heading',
+      ],
+      'title_tag' => [
+        [$this, 'assertTitleTag'],
+      ],
+      'background' => [
+        [$this, 'assertBackground'],
+      ],
       'badges' => [
         [$this, 'assertBadgesElements'],
       ],
-      'title' => [
-        [$this, 'assertElementText'],
-        '.card-title',
+      'links' => [
+        [$this, 'assertLinks'],
       ],
-      'description' => [
-        [$this, 'assertDescription'],
+      'content' => [
+        [$this, 'assertElementText'],
+        '.banner-content',
       ],
       'meta' => [
         [$this, 'assertMeta'],
+      ],
+      'action_bar' => [
+        [$this, 'assertActionBar'],
       ],
     ];
   }
@@ -46,6 +59,58 @@ class ContentBannerAssert extends BasePatternAssert {
   }
 
   /**
+   * Asserts the rendered html of the action-bar.
+   *
+   * @param string|null $expected
+   *   The expected value.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertActionBar($expected, Crawler $crawler): void {
+    if (is_null($expected)) {
+      $this->assertElementNotExists('.bcl-content-banner > .container > *:last-child', $crawler);
+      return;
+    }
+    $this->assertElementExists('.bcl-content-banner > .container > *:last-child', $crawler);
+    $element = $crawler->filter('.bcl-content-banner > .container > *:last-child');
+    self::assertEquals($expected, $element->outerHtml());
+  }
+
+  /**
+   * Asserts the background class of the content banner.
+   *
+   * @param string $expected
+   *   The expected background class.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The dom crawler.
+   */
+  protected function assertBackground(string $expected, Crawler $crawler): void {
+    $element = $crawler->filter('.bcl-content-banner');
+
+    if ($expected === 'white') {
+      self::assertStringContainsString('bg-white', $element->attr('class'));
+    }
+    elseif ($expected === 'gray') {
+      self::assertStringContainsString('bg-lighter', $element->attr('class'));
+    }
+    else {
+      self::assertStringContainsString('bg-transparent', $element->attr('class'));
+    }
+  }
+
+  /**
+   * Checks the tag used for the title.
+   *
+   * @param string $expected
+   *   The expected tag.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertTitleTag(string $expected, Crawler $crawler): void {
+    $this->assertElementExists($expected . '.card-title.bcl-heading', $crawler);
+  }
+
+  /**
    * Asserts the content banner description.
    *
    * @param string $text
@@ -57,6 +122,26 @@ class ContentBannerAssert extends BasePatternAssert {
     $element = $crawler->filter('.card-body');
     self::assertCount(1, $element);
     self::assertStringContainsString($text, $element->text());
+  }
+
+  /**
+   * Asserts the content banner links.
+   *
+   * @param array $expected
+   *   The expected links.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The dom crawler.
+   */
+  protected function assertLinks(array $expected, Crawler $crawler): void {
+    $actualLinks = $crawler->filter('.card-body div:last-child > div')->each(function (Crawler $element) {
+      return [
+        'label' => $element->filter('a')->text(),
+        'path' => $element->filter('a')->attr('href'),
+      ];
+    });
+
+    // Assert that the actual links match the expected links.
+    $this->assertEquals($expected, $actualLinks);
   }
 
   /**
--- a/tests/src/PatternAssertion/ContentBannerAssert.php
+++ b/tests/src/PatternAssertion/ContentBannerAssert.php
@@ -41,7 +41,7 @@ class ContentBannerAssert extends BasePatternAssert {
       ],
       'content' => [
         [$this, 'assertElementText'],
-        '.card-text',
+        '.card-content',
       ],
       'meta' => [
         [$this, 'assertMeta'],
--- a/tests/src/PatternAssertion/ContentBannerAssert.php
+++ b/tests/src/PatternAssertion/ContentBannerAssert.php
@@ -90,15 +90,7 @@ class ContentBannerAssert extends BasePatternAssert {
   protected function assertBackground(string $expected, Crawler $crawler): void {
     $element = $crawler->filter('.bcl-content-banner');
 
-    if ($expected === 'white') {
-      self::assertStringContainsString('bg-white', $element->attr('class'));
-    }
-    elseif ($expected === 'gray') {
-      self::assertStringContainsString('bg-lighter', $element->attr('class'));
-    }
-    else {
-      self::assertStringContainsString('bg-transparent', $element->attr('class'));
-    }
+    self::assertStringContainsString('bg-' . $expected, $element->attr('class'));
   }
 
   /**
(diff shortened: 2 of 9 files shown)


tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertImageSize(string $expected, Crawler $crawler): void {
+    $element = $crawler->filter('.bcl-card-start-col');
+
+    if ($expected === 'xl') {
+      self::assertStringContainsString('bcl-size-extra-large', $element->attr('class'));
+    }
+    elseif ($expected === 'md') {
+      self::assertStringContainsString('bg-lighter', $element->attr('class'));
```
AaronGilMartinez: I think this is for image size and we expect other class

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+    }
+    else {
+      // Check that the element has only one class: bcl-card-start-col.
+      $classes = explode(' ', $element->attr('class'));
+      self::assertCount(1, $classes);
+      self::assertEquals('bcl-card-start-col', $classes[0]);
```
AaronGilMartinez: Default class check for 'lg' is missing

tests/src/PatternAssertion/ContentBannerAssert.php
AaronGilMartinez: We have two cases in the test fixture yml with the same image size 'xl', could we have one of them with 'md' to cover all possible options, please?

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+    $element = $crawler->filter('.bcl-content-banner');
+
+    if ($expected === 'white') {
+      self::assertStringContainsString('bg-white', $element->attr('class'));
+    }
+    elseif ($expected === 'gray') {
+      self::assertStringContainsString('bg-lighter', $element->attr('class'));
+    }
+    else {
+      self::assertStringContainsString('bg-transparent', $element->attr('class'));
```
AaronGilMartinez: We check values white and gray, but not default transparent, maybe remove one of the values?

tests/src/PatternAssertion/InPageNavigationAssert.php
```diff
@@ -0,0 +1,60 @@
...
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
+    $crawler = new Crawler($html);
+    $page_header = $crawler->filter('body');
```
AaronGilMartinez: I think this variable name is some inherited code from other test. 
Also, should this tag be checked here or  should be  `ul.nav-pills`?
And check that there is only one element like this.

tests/src/PatternAssertion/InPageNavigationAssert.php
```diff
@@ -0,0 +1,60 @@
...
+  /**
+   * Asserts the in-page-navigation links list.
+   *
+   * @param array|null $expected
+   *   The expected description values.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertList($expected, Crawler $crawler): void {
+    $this->assertElementExists('ul.nav-pills', $crawler);
```
AaronGilMartinez: If we check this in the base element assert we don't need it here.

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,168 @@
...
+        [$this, 'assertItems'],
+      ],
+    ];
+  }
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function assertBaseElements(string $html, string $variant): void {
+    $this->assertElementExists('body > .bcl-description-list', new Crawler($html));
```
AaronGilMartinez: Can we add a check to see that it is the only description list?

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,168 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_bootstrap_theme\PatternAssertion;
+
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Assertions for the field list pattern.
```
AaronGilMartinez:
```suggestion
 * Assertions for the description list pattern.
```

AaronGilMartinez (changes requested):

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: sorry, i dont really understand this point. We test all the cases possible.
    - background: (string) (default: 'white')
      options: [ white, gray ]
      
      there are only 2 options. for transparent you dont add this parameter which is covered.

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: done. that cases should be for lg which i redone.

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: done.

tests/src/PatternAssertion/InPageNavigationAssert.php
tibi2303: done.

tests/src/PatternAssertion/InPageNavigationAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -41,7 +41,7 @@
     'path': image.src,
     'alt': image.alt,
   } : {},
-  'image_size': image_size ?? 'lg',
+  'image_size': image_size|default('lg'),
```
drishu: this change is not BC, `??` and default are not identical in behaviour

templates/patterns/content_banner/pattern-content-banner.html.twig
```diff
@@ -7,9 +7,9 @@
 {% apply spaceless %}
 {# Render the 'meta' items into the content variable. #}
 {% set _content %}
-  {% set content_attributes = create_attribute().addClass('mt-4') %}
+  {% set content_attributes = create_attribute().addClass(['mt-4', 'banner-content']) %}
   {% if meta is not empty %}
-    {% set content_attributes = create_attribute() %}
+    {% set content_attributes = create_attribute().addClass('banner-content') %}
```
drishu: Semantic classes are definitely the way we want to move forward, but `banner-content` is quite confusing, lets rather name it `card-body` (logic being that this gets injected into the card

tests/src/PatternAssertion/ContentBannerAssert.php
drishu: if the default is white, then what value do I need to set for it to be transparent?

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+  protected function assertBackground(string $expected, Crawler $crawler): void {
+    $element = $crawler->filter('.bcl-content-banner');
+
+    if ($expected === 'white') {
+      self::assertStringContainsString('bg-white', $element->attr('class'));
+    }
+    elseif ($expected === 'gray') {
+      self::assertStringContainsString('bg-lighter', $element->attr('class'));
+    }
+    else {
```
drishu: dont think you need to cast this one, we can make it more flexible like 
`self::assertStringContainsString('bg-' . $expected, $element->attr('class'));`

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The DomCrawler where to check the element.
+   */
+  protected function assertImageSize(string $expected, Crawler $crawler): void {
+    $element = $crawler->filter('.bcl-card-start-col');
+
+    if ($expected === 'xl') {
+      self::assertStringContainsString('bcl-size-extra-large', $element->attr('class'));
+    }
+    elseif ($expected === 'lg') {
```
drishu: it can also be `md` or other sizes, so maybe map the classes with php match and then do just one assert with that found class name

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+
+  /**
+   * Asserts the content banner description.
+   *
+   * @param string $text
+   *   The expected description.
+   * @param \Symfony\Component\DomCrawler\Crawler $crawler
+   *   The dom crawler.
+   */
+  protected function assertDescription(string $text, Crawler $crawler): void {
```
drishu: my editor says its not used

tests/src/PatternAssertion/ContentBannerAssert.php
```diff
@@ -0,0 +1,191 @@
...
+        '.card-title.bcl-heading',
+      ],
+      'title_tag' => [
+        [$this, 'assertTitleTag'],
+      ],
+      'background' => [
+        [$this, 'assertBackground'],
+      ],
+      'badges' => [
+        [$this, 'assertBadgesElements'],
```
drishu, line 37: this `assertBadgesElements` is actually not defined, dangrous that it fails silently, if you have time you should look also into that, if not, make a separate ticket for it.

Same for ./tests/src/PatternAssertion/CardPatternAssert.php

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,170 @@
...
+   * @param array $labels
+   *   An array to store labels.
+   * @param array $values
+   *   An array to store values.
+   * @param array $dt_icons
+   *   An array to store icons for <dt> elements.
+   * @param array $dd_icons
+   *   An array to store icons for <dd> elements.
+   */
+  private function extractLabelsValuesAndIcons(array $expected_items, array &$labels, array &$values, array &$dt_icons, array &$dd_icons): void {
```
drishu: this violates the single responsibility principle

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,170 @@
...
+    foreach ($expected_items as $item) {
+      $this->extractLabelsAndIcons($item['term'] ?? NULL, $labels, $dt_icons);
+      $this->extractLabelsAndIcons($item['definition'] ?? NULL, $values, $dd_icons);
```
drishu: also this violates the single responsibility principle, and its naming is deceiving, it seems you can use it also for values, not just icons and labels

drishu (changes requested):

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: indeed, refactored.

templates/patterns/content_banner/pattern-content-banner.html.twig
tibi2303: ok, i've rollback.

templates/patterns/content_banner/pattern-content-banner.html.twig
tibi2303: added it as card-text.. we already have card-body in that structure.. this is an inner child

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: done.

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: done.

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: removed.

tests/src/PatternAssertion/ContentBannerAssert.php
tibi2303: done. added in base pattern.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: refactored.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: refactored.

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   *   The item to extract label from.
+   *
+   * @return array
+   *   The extracted labels.
+   */
+  private function extractLabel($item): array {
+    $labels = [];
+    if ($item === NULL) {
+      return $labels;
+    }
```
drishu: you already checked if its set in extractLabels()

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   * @param string $key
+   *   The key to extract labels for ('term' or 'definition').
+   *
+   * @return array
+   *   The extracted labels.
+   */
+  private function extractLabels(array $expected_items, string $key): array {
+    $labels = [];
+    foreach ($expected_items as $item) {
+      if (isset($item[$key])) {
```
drishu: Lets flatten the code a bit and do
```
if (!isset($item[$key])) {
  continue;
}
```

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   * @return array
+   *   The extracted labels.
+   */
+  private function extractLabel($item): array {
+    $labels = [];
+    if ($item === NULL) {
+      return $labels;
+    }
+
+    if (is_array($item)) {
```
drishu: If we move this if into extractLabel then we could have 2 functions, one for arrays one for strings

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+      foreach ($item as $subItem) {
+        if (isset($subItem['label'])) {
+          $labels[] = is_array($subItem['label']) ? strip_tags($subItem['label'][0]['#markup'] ?? $subItem['label'][0]) : strip_tags($subItem['label']);
```
drishu: this is a bit hard to ready, first we could move the result of this expression into a separate variable above, and do stripe tags just once
`$labels[] = strip_tags($label)`

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   */
+  private function extractLabel($item): array {
+    $labels = [];
+    if ($item === NULL) {
+      return $labels;
+    }
+
+    if (is_array($item)) {
+      foreach ($item as $subItem) {
+        if (isset($subItem['label'])) {
```
drishu: there is no else, so it would remain silent if the structure fails, lets not add the if so it crashes and burns

templates/patterns/content_banner/pattern-content-banner.html.twig
drishu: Let's name it card-content as text suggests there is no markup

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   *   The key to extract labels for ('term' or 'definition').
+   *
+   * @return array
+   *   The extracted labels.
+   */
+  private function extractLabels(array $expected_items, string $key): array {
+    $labels = [];
+    foreach ($expected_items as $item) {
+      if (isset($item[$key])) {
+        $labels = array_merge($labels, $this->extractLabel($item[$key]));
```
drishu: here you can add the if else, so if its not array or string it should fail

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,202 @@
...
+   *   The key to extract icons for ('term' or 'definition').
+   *
+   * @return array
+   *   The extracted icons.
+   */
+  private function extractIcons(array $expected_items, string $key): array {
+    $icons = [];
+    foreach ($expected_items as $item) {
+      if (isset($item[$key])) {
+        $icons = array_merge($icons, $this->extractIcon($item[$key]));
```
drishu: same comments as for the other method pairs

drishu (changes requested):

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done.

tests/src/PatternAssertion/DescriptionListAssert.php
tibi2303: done

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,228 @@
...
+        $labels = array_merge($labels, $this->extractLabelFromString($item[$key]));
+      }
+    }
+    return $labels;
+  }
+
+  /**
+   * Extracts label from an array of item.
+   *
+   * @param mixed $item
```
drishu:
```suggestion
   * @param array $item
```

tests/src/PatternAssertion/DescriptionListAssert.php
```diff
@@ -0,0 +1,228 @@
...
+      $labels[] = strip_tags($label);
+    }
+
+    return $labels;
+  }
+
+  /**
+   * Extracts label from an item string.
+   *
+   * @param mixed $item
```
drishu:
```suggestion
   * @param string $item
```

drishu (changes requested):

drishu (dismissed):

AaronGilMartinez (dismissed):

AaronGilMartinez (approved):

drishu (approved):

PR #396, opened 2024-05-17
OEL-1863: Changed selector from FilePattern.

tibi2303:
Jira issue(s):

### DIFF
--- a/tests/src/PatternAssertion/FilePatternAssert.php
+++ b/tests/src/PatternAssertion/FilePatternAssert.php
@@ -43,8 +43,7 @@ class FilePatternAssert extends BasePatternAssert {
     $crawler = new Crawler($html);
     // Only one icon should be rendered. The type will be asserted in the file
     // assert.
-    $this->assertElementExists('svg.icon--2xl', $crawler);
-    $this->assertElementExists('svg.icon--file', $crawler);
+    $this->assertElementExists('svg.icon--file.icon--2xl', $crawler);
     // Only one wrapper element is present.
     $this->assertCount(1, $crawler->filter('.bcl-file-container'));
     // Only one wrapper element is present within bcl-file-container.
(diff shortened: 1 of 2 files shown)


tests/src/PatternAssertion/FilePatternAssert.php
```diff
@@ -53,7 +54,7 @@ protected function assertBaseElements(string $html, string $variant): void {
    *   The crawler.
    */
   protected function assertFile(array $expected, Crawler $crawler): void {
-    $file = $crawler->filter('body > div.mt-4 > div.border.rounded > div:first-child');
+    $file = $crawler->filter('body > .bcl-file-container > .bcl-file.border.rounded > div:first-child');
```
AaronGilMartinez: The change with the wrapper selector is good, although I think the ticket is asking to remove this direct child selector to allow the use of this assert in the whitelabel person test like content banner assert here: https://github.com/openeuropa/oe_whitelabel/blob/cc8f899dd188e24d3e4b4c83a15012339282c4a1/tests/src/Functional/PublicationContentRenderTest.php#L46

Can you check, please?
If it's the case there may be other places in this assert class with direct body child selector.

AaronGilMartinez (changes requested):

tests/src/PatternAssertion/FilePatternAssert.php
tibi2303: you are right. I removed the direct body > selectors.

AaronGilMartinez (changes requested):
Changes are ok, only needs a rebase

AaronGilMartinez (dismissed):

tests/src/PatternAssertion/FilePatternAssert.php
```diff
@@ -46,8 +46,9 @@ protected function assertBaseElements(string $html, string $variant): void {
     $this->assertElementExists('svg.icon--2xl', $crawler);
     $this->assertElementExists('svg.icon--file', $crawler);
```
drishu: I think we should change this, lets just check the presence of the svg, or even remove these lines (See what makes sense) as we have the IconPatternAssert test

drishu (changes requested):

tests/src/PatternAssertion/FilePatternAssert.php
tibi2303: image

it's always visible inside the file but indeed we can make it 1 line.

drishu (approved):

AaronGilMartinez (approved):

PR #397, opened 2024-06-06
OEL-2743: Theme views table.

### LINKED ISSUE #312: Theme views table
Recently you have themed the tables https://github.com/openeuropa/oe_bootstrap_theme/pull/290 using BCL. Thanks!
It would be nice to also consider theming the views table.

AaronGilMartinez:
Jira issue(s):


oe_bootstrap_theme.theme
```diff
@@ -56,3 +56,29 @@ function oe_bootstrap_theme_preprocess_table__bootstrap__responsive(&$variables)
 
   $variables['wrapper_attributes'] = new Attribute(['class' => $class]);
 }
+
+/**
+ * Implements hook_preprocess_HOOK() for views-view-table.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_views_view_table(&$variables) {
```
drishu: Let's go with the same approach as we did with the table templates:

- implement `oe_bootstrap_theme_theme_suggestions_views_view_table_alter`
- there we add `views_view_table__bootstrap` and `views_view_table__bootstrap__responsive`.
- then  implement `oe_bootstrap_theme_preprocess_views_view_table__bootstrap__responsive`
- template `views-view-table--bootstrap.html.twig`
- and `views-view-table--bootstrap--responsive.html.twig` extending the first one

tests/src/Kernel/ViewsTableTest.php
```diff
@@ -0,0 +1,184 @@
...
+    'views_test_data',
+  ];
+
+  /**
+   * {@inheritdoc}
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Configuration and schemas needed for view test.
```
drishu: instead of this comment and the one from line 38, rather explain why we do all this code ex: `// Here we set up tralalal` and leave a @see reference to ViewsKernelTestBase::setUpFixtures()

tests/src/Kernel/ViewsTableTest.php
```diff
@@ -0,0 +1,184 @@
...
+    $query->execute();
+
+    // Enable the test table view.
+    $storage = \Drupal::entityTypeManager()->getStorage('view');
+    $config_dir = \Drupal::service('extension.list.module')->getPath('views_test_config') . '/test_views';
+    $file_storage = new FileStorage($config_dir);
+    $storage->create($file_storage->read('views.view.test_table'))->save();
+
+    // Enable the theme.
+    $this->container->get('theme_installer')->install(['oe_bootstrap_theme']);
```
drishu: Isn't the theme already enabled in the parent::setUp ?

drishu (changes requested):

tests/src/Kernel/ViewsTableTest.php
AaronGilMartinez: Indeed, I was reinstalling system configuration and causing issues.

tests/src/Kernel/ViewsTableTest.php
AaronGilMartinez: Ok, this way explains why the code it's there and that I'm following core

oe_bootstrap_theme.theme
AaronGilMartinez: Thanks, this saved time

oe_bootstrap_theme.theme
```diff
@@ -48,6 +68,20 @@ function oe_bootstrap_theme_preprocess(&$variables) {
...
+
+/**
+ * Implements hook_preprocess_HOOK() for views-view-table--bootstrap--responsive.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_views_view_table__bootstrap__responsive(&$variables) {
+  _oe_bootstrap_theme_bootstrap_responsive_table($variables);
+}
+
+/**
+ * Function that applies responsiveness to Bootstrap tables.
```
drishu, line 82: Since this is not a drupal hook you need to add parameters

oe_bootstrap_theme.theme
```diff
@@ -48,6 +68,20 @@ function oe_bootstrap_theme_preprocess(&$variables) {
...
+ */
+function _oe_bootstrap_theme_bootstrap_responsive_table(&$variables) {
```
drishu: add return type

templates/patterns/offcanvas/pattern-offcanvas.html.twig
```diff
@@ -10,8 +10,8 @@
   'placement': placement,
   'id': offcanvas_id,
   'body': body,
-  'with_body_scroll': body_scroll,
-  'with_backdrop': backdrop,
+  'with_body_scroll': true,
+  'with_backdrop': true,
```
drishu: why these changes? seems unrelated and these variables exist in the pattern definition so its not ok to change them

drishu (changes requested):

oe_bootstrap_theme.theme
AaronGilMartinez: True

oe_bootstrap_theme.theme
AaronGilMartinez: Ok

templates/patterns/offcanvas/pattern-offcanvas.html.twig
AaronGilMartinez: Added by mistake

drishu (approved):

intelektron (approved):

PR #412, opened 2024-07-30
ELA-1118: Add schema for the filter settings

### LINKED ISSUE #411: Drupal 10.3: 'responsive' is not a supported key.
### The problem

Drupal 10.3 + oe_bootstrap_theme 

Error message:
`'responsive' is not a supported key.`

While saving an edit page of any text format using CKEditor 5, ex. http://web:8080/web/en/admin/config/content/formats/manage/full_html

This issue doesn't happen with Drupal 10.2 (downgrade fixes the issue).

Also, when removing below file, the problem is gone: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/modules/oe_bootstrap_theme_helper/src/Plugin/Filter/BootstrapTable.php

piotrsmykaj:
Jira issue(s):
- [ELA-1118](https://webgate.ec.europa.eu/CITnet/jira/browse/ELA-1118)

GitHub issue:
- https://github.com/openeuropa/oe_bootstrap_theme/issues/411

### DIFF
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,4 +1,3 @@
-version: '2'
 services:
   web:
     image: fpfis/httpd-php-dev:8.2
--- a/docker-compose.yml
+++ b/docker-compose.yml
@@ -1,3 +1,4 @@
+version: '2'
 services:
   web:
     image: fpfis/httpd-php-dev:8.2
--- a/modules/oe_bootstrap_theme_helper/src/Plugin/Filter/BootstrapTable.php
+++ b/modules/oe_bootstrap_theme_helper/src/Plugin/Filter/BootstrapTable.php
@@ -6,23 +6,21 @@ namespace Drupal\oe_bootstrap_theme_helper\Plugin\Filter;
 
 use Drupal\Component\Utility\Html;
 use Drupal\Core\Form\FormStateInterface;
-use Drupal\Core\StringTranslation\TranslatableMarkup;
-use Drupal\filter\Attribute\Filter;
 use Drupal\filter\FilterProcessResult;
 use Drupal\filter\Plugin\FilterBase;
-use Drupal\filter\Plugin\FilterInterface;
 
 /**
  * Provides a filter to add Bootstrap styling to tables.
+ *
+ * @Filter(
+ *   id = "oe_bootstrap_theme_table",
+ *   title = @Translation("Bootstrap tables"),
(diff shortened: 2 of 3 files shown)


docker-compose.yml
```diff
@@ -1,4 +1,3 @@
-version: '2'
```
AaronGilMartinez: This change doesn't seem related with the issue. Why this?

modules/oe_bootstrap_theme_helper/src/Plugin/Filter/BootstrapTable.php
```diff
@@ -6,21 +6,23 @@
...
- * )
  */
+#[Filter(
+  id: "oe_bootstrap_theme_table",
+  title: new TranslatableMarkup("Bootstrap tables"),
+  type: FilterInterface::TYPE_TRANSFORM_IRREVERSIBLE,
+  settings: [
+    "responsive" => "",
+  ],
+)]
```
AaronGilMartinez: Core introduced discovery for filter Attributes on 10.3.0 https://www.drupal.org/node/3229001
This change won't support 10.2.x

AaronGilMartinez (changes requested):

docker-compose.yml
piotrsmykaj: while working on this fix I received local warnings about a deprecated version parameter

modules/oe_bootstrap_theme_helper/src/Plugin/Filter/BootstrapTable.php
piotrsmykaj: yes good point, in this case it needs to be reverted

docker-compose.yml
AaronGilMartinez: The "version" is used to declare file format it's only used for Compose v1, Compose v2 will display a warning and this setting be ignored. I know the intention is good, but this is out of scope and it may be used in some cases, lets remove this change.

AaronGilMartinez (dismissed):

julien-:
Take a look in https://www.drupal.org/project/drupal/releases/10.3.2

> Text formats could not be saved due to config translation validation new in Drupal 10.3. The LogicException has been replaced with a E_USER_DEPRECATED error.

brummbar (approved):

PR #417, opened 2024-08-08
OEL-1970: Removed empty sass files.

tibi2303:
Jira issue(s):


brummbar:
Closing as we will leave the files as example for child themes.

PR #420, opened 2024-08-13
OEL-3024: Added tests for new class on pagination.

tibi2303:
Jira issue(s):

### DIFF
--- a/tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml
+++ b/tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml
@@ -6,6 +6,7 @@ pagination:
       settings:
         size: 'sm'
         alignment: 'center'
+        list_attributes: 'class="list-class" id="list-id" data-list-attribute="list-attr-value"'
       enable_first_last_icons: false
       first:
         path: '#first'
@@ -25,7 +26,6 @@ pagination:
           active: true
         - path: '#2'
           label: '2'
-      list_attributes: 'class="list-class" id="list-id" data-list-attribute="list-attr-value"'
   assertions:
     pattern:
       class: Drupal\Tests\oe_bootstrap_theme\PatternAssertion\PaginationPatternAssert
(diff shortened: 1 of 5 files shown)


tests/src/PatternAssertion/PaginationPatternAssert.php
```diff
@@ -25,6 +25,10 @@ protected function getAssertions(string $variant): array {
       'size' => [
         [$this, 'assertSize'],
       ],
+      'list_attributes' => [
+        [$this, 'assertElementAttributes'],
+        '.pagination.js-pager__items',
```
drishu: hardcoding this is wrong, we set what is expected, we need to assert against that

drishu (changes requested):

tests/src/PatternAssertion/PaginationPatternAssert.php
tibi2303: changed the selector.

drishu (dismissed):

tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml
```diff
@@ -6,6 +6,7 @@ pagination:
       settings:
         size: 'sm'
         alignment: 'center'
+        list_attributes: 'class="list-class" id="list-id" data-list-attribute="list-attr-value"'
```
brummbar: As I said in previous reviews, do not change the default test. In here we should just test the default values.
Add a new one, with the extra attributes.

brummbar (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/pagination.yml
tibi2303: did it in a separate test.

drishu (approved):

brummbar (approved):

PR #422, opened 2024-08-19
OEL-2283: Added overridable icon_path.

tibi2303:
Jira issue(s):


modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,108 @@
...
+    $variables = [];
+
+    // Call the preprocess function manually.
+    oe_bootstrap_theme_preprocess($variables);
+
+    // Extract the path set in the preprocess function.
+    $actualIconPath = $variables['bcl_icon_path'];
+
+    // Check if the paths match.
+    $this->assertEquals($expectedIconPath, $actualIconPath);
```
AaronGilMartinez: Not sure whats the expected value here, we have one test with one result.
There is a coalesce operator on line 58, let's make it static value to know whats the expected value, remove coalesce.
1. Assert the theme bcl_icon_path has a value like 'assets/icons/bcl-default-icons-test.svg'
2. Assert that the value is the expected value.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,108 @@
...
+    $themeHandler = \Drupal::service('theme_handler');
+
+    // Switch to the base theme.
+    $this->config('system.theme')->set('default', 'oe_bootstrap_theme')->save();
+    // Retrieve the active theme.
+    $activeTheme = $this->themeManager->getActiveTheme();
+
+    // Access the bcl_icon_path from the theme info file.
+    $themeInfo = $activeTheme->getExtension()->info;
+    $bclIconPath = $themeInfo['openeuropa']['bootstrap_component_library']['bcl_icon_path'] ?? 'assets/icons/bcl-default-icons.svg';
```
AaronGilMartinez: Same here, we don't know if the value for bcl_icon_path is null or not.
1. Assert value of setting, if is not set check that.
2. Assert expected output, fallback.

For me fallback should be tested from child theme as is done in override, but no value is set.

You can do the same as for components_location, and use a oe_bootstrap_theme_test_subtheme2 without value.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
AaronGilMartinez, line 1: If the variable is managed by theme I think the test should be placed in theme tests, not module. I would add separate tests theme too, with minimum info.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,108 @@
...
+    // Check if the paths match.
+    $this->assertEquals($expectedIconPath, $actualIconPath);
+  }
+
+  /**
+   * Tests the fallback when the bcl_icon_path is not set in the subtheme.
+   */
+  public function testBclIconPathFallback(): void {
+    // Use the ThemeHandler service to retrieve the base theme.
+    $themeHandler = \Drupal::service('theme_handler');
```
AaronGilMartinez: This variable is not used.

oe_bootstrap_theme.theme
```diff
@@ -61,7 +61,15 @@ function oe_bootstrap_theme_theme_suggestions_views_view_table_alter(array &$sug
  * Implements hook_preprocess().
  */
 function oe_bootstrap_theme_preprocess(&$variables) {
-  $variables['bcl_icon_path'] = base_path() . \Drupal::service('extension.list.theme')->getPath('oe_bootstrap_theme') . '/assets/icons/bcl-default-icons.svg';
+  // Get the active theme and its path.
+  $theme = \Drupal::theme()->getActiveTheme();
+  $theme_path = \Drupal::service('extension.list.theme')->getPath($theme->getName());
```
AaronGilMartinez:
```suggestion
  $theme_path = $theme->getPath();
```

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,108 @@
...
+ */
+class BclIconPathOverrideTest extends KernelTestBase {
+
+  /**
+   * {@inheritdoc}
+   */
+  protected static $modules = [
+    'system',
+    'theme_test',
+    'oe_bootstrap_theme_helper',
```
AaronGilMartinez: Not needed helper module, the theme does the preprocess.
Module theme_test can be removed too.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,108 @@
...
+    // Initialize the theme manager service.
+    $this->themeManager = $this->container->get('theme.manager');
+
+    // Install the theme. Subsequently, this installs also the base themes.
+    $this->container->get('theme_installer')->install(['oe_bootstrap_theme', 'oe_bootstrap_theme_test_subtheme1']);
+    $this->config('system.theme')->set('default', 'oe_bootstrap_theme_test_subtheme1')->save();
+
+    // Manually include the .theme file of the base theme.
+    $themePath = \Drupal::service('extension.list.theme')->getPath('oe_bootstrap_theme');
+    require_once $themePath . '/' . 'oe_bootstrap_theme.theme';
```
AaronGilMartinez: Not needed.

oe_bootstrap_theme.theme
AaronGilMartinez: Apply everywhere, please.

AaronGilMartinez (changes requested):

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

oe_bootstrap_theme.theme
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

modules/oe_bootstrap_theme_helper/tests/themes/oe_bootstrap_theme_test_subtheme1/oe_bootstrap_theme_test_subtheme1.info.yml
```diff
@@ -6,3 +6,4 @@ core_version_requirement: ^10
 openeuropa:
   bootstrap_component_library:
     components_location: components/templates
+    bcl_icon_path: 'assets/icons/bcl-default-icons-test.svg'
```
AaronGilMartinez: We don't need this here if we create test themes.

oe_bootstrap_theme.theme
```diff
@@ -61,7 +61,15 @@ function oe_bootstrap_theme_theme_suggestions_views_view_table_alter(array &$sug
  * Implements hook_preprocess().
  */
 function oe_bootstrap_theme_preprocess(&$variables) {
-  $variables['bcl_icon_path'] = base_path() . \Drupal::service('extension.list.theme')->getPath('oe_bootstrap_theme') . '/assets/icons/bcl-default-icons.svg';
+  // Get the active theme and its path.
+  $theme = \Drupal::theme()->getActiveTheme();
+  $theme_path = $theme->getPath($theme->getName());
```
AaronGilMartinez: This method doesn't accept parameters, the test are failing which is good because is telling us something is not ok.

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    // Retrieve the active theme.
+    $theme = $this->themeManager->getActiveTheme();
+
+    // Access the bcl_icon_path from the theme info file.
+    $themeInfo = $theme->getExtension()->info;
+
+    // Optionally, check if the variable is empty if it exists.
+    $bclIconPath = $themeInfo['openeuropa']['bootstrap_component_library']['bcl_icon_path'] ?? NULL;
+    $this->assertEmpty($bclIconPath);
```
AaronGilMartinez:
```suggestion
    $this->assertNull($bclIconPath);
```

AaronGilMartinez (changes requested):

modules/oe_bootstrap_theme_helper/tests/themes/oe_bootstrap_theme_test_subtheme1/oe_bootstrap_theme_test_subtheme1.info.yml
tibi2303: done.

oe_bootstrap_theme.theme
tibi2303: done.

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

AaronGilMartinez (dismissed):

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+  protected static $modules = [
+    'system',
+  ];
+
+  /**
+   * The theme manager service.
+   *
+   * @var \Drupal\Core\Theme\ThemeManagerInterface
+   */
+  protected ThemeManagerInterface $themeManager;
```
brummbar: We don't set services as properties, unless very specific scenarios.

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Set up the test environment.
+   */
+  protected function setUp(): void {
+    parent::setUp();
+
+    // Initialize the theme manager service.
```
brummbar: We are missing the file_scan_ignore_directories behaviour as per \Drupal\Tests\oe_bootstrap_theme_helper\Kernel\BclComponentsOverrideTest::setUp().

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+    $this->themeManager = $this->container->get('theme.manager');
+
+    // Install the theme. Subsequently, this installs also the base themes.
+    $this->container->get('theme_installer')->install(['oe_bootstrap_theme', 'oe_bootstrap_theme_test_subtheme']);
+    $this->config('system.theme')->set('default', 'oe_bootstrap_theme_test_subtheme')->save();
```
brummbar: There's no need to do this here.

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+  /**
+   * Tests that the bcl_icon_path is correctly overridden in the subtheme.
+   */
+  public function testBclIconPathOverride(): void {
+    // Retrieve the active theme.
+    $theme = $this->themeManager->getActiveTheme();
+
+    // Access the bcl_icon_path from the theme info file.
+    $themeInfo = $theme->getExtension()->info;
```
brummbar: snake_case, no CamelCase.

tests/theme/oe_bootstrap_theme_test_subtheme/oe_bootstrap_theme_test_subtheme.info.yml
```diff
@@ -0,0 +1,8 @@
+name: OpenEuropa Bootstrap Theme Test Subtheme
```
brummbar: Didn't we say about using the already existing test themes?

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,104 @@
...
+
+    // Retrieve the active theme.
+    $theme = $this->themeManager->getActiveTheme();
+
+    // Access the bcl_icon_path from the theme info file.
+    $themeInfo = $theme->getExtension()->info;
+
+    // Optionally, check if the variable is empty if it exists.
+    $bclIconPath = $themeInfo['openeuropa']['bootstrap_component_library']['bcl_icon_path'] ?? NULL;
+    $this->assertNull($bclIconPath);
```
brummbar: We don't need to run these checks, we don't need to test our test data. We only need to run the preprocess and compare it with a static value that we know it's in the theme info file, like \Drupal\Tests\oe_bootstrap_theme_helper\Kernel\BclComponentsOverrideTest::bclComponentsOverridingTestCasesProvider.

brummbar (changes requested):
Let's use a dataprovider with 3 scenarios like \Drupal\Tests\oe_bootstrap_theme_helper\Kernel\BclComponentsOverrideTest::bclComponentsOverridingTestCasesProvider.

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: removed it

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: added

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: yep, refactored the tests with data provider

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: redone it

tests/theme/oe_bootstrap_theme_test_subtheme/oe_bootstrap_theme_test_subtheme.info.yml
tibi2303: yep

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: refactored it

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,92 @@
...
+    ];
+    new Settings($settings);
+  }
+
+  /**
+   * Tests that the bcl_icon_path is correctly handled in the theme.
+   *
+   * @param string $theme
+   *   The theme to set as default.
+   * @param string|null $expected_icon_path
```
AaronGilMartinez:
```suggestion
   * @param string $expected_icon_path
```

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,92 @@
...
+   * @dataProvider bclIconPathTestCasesProvider
+   */
+  public function testBclIconPath(string $theme, string $expected_icon_path): void {
+    // Install the theme. Subsequently, this installs also the base themes.
+    $this->container->get('theme_installer')->install([$theme]);
+    $this->config('system.theme')->set('default', $theme)->save();
+
+    // Retrieve the active theme.
+    $themeManager = $this->container->get('theme.manager');
+    $active_theme = $themeManager->getActiveTheme();
```
AaronGilMartinez:
```suggestion
    \Drupal::theme()->getActiveTheme();
```

AaronGilMartinez (changes requested):
Small changes

AaronGilMartinez (dismissed):

tests/src/Kernel/BclIconPathOverrideTest.php
```diff
@@ -0,0 +1,92 @@
...
+   */
+  public function bclIconPathTestCasesProvider(): array {
+    return [
+      'subtheme theme with overridden path' => [
+        'oe_bootstrap_theme_test_subtheme1',
+        '/themes/custom/oe_bootstrap_theme/modules/oe_bootstrap_theme_helper/tests/themes/oe_bootstrap_theme_test_subtheme1/assets/icons/bcl-default-icons-test.svg',
+      ],
+      'subtheme theme with fallback path' => [
+        'oe_bootstrap_theme_test_subtheme2',
+        '/themes/custom/oe_bootstrap_theme/modules/oe_bootstrap_theme_helper/tests/themes/oe_bootstrap_theme_test_subtheme2/assets/icons/bcl-default-icons.svg',
```
brummbar: Now that we have these tests, it outlines that the current implementation is flawed.
When a theme doesn't declare the value, it should use the value of the parent theme, if present. All the way up the chain until hitting oe_bootstrap_theme.
We don't want to force subthemes to copy the svg in their folder if no override is needed.

brummbar (changes requested):

tests/src/Kernel/BclIconPathOverrideTest.php
tibi2303: done.

brummbar (approved):

AaronGilMartinez (approved):

PR #425, opened 2024-08-28
OEL-3019: Improved spacing between content in card.

### LINKED ISSUE #386: Card pattern meta and text add extra space to the bottom
When not passing content to the card pattern (any variant) but passing meta, then we have an extra space on the bottom. This is due to the fact that `my-3` is adding the space to the bottom regardless there is content passed or not. See: https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/card/pattern-card--variant-search.html.twig#L11

Same issue when not passing content to the pattern but passing text. This one is due to `mb-2-5` in https://github.com/openeuropa/oe_bootstrap_theme/blob/1.x/templates/patterns/card/pattern-card--variant-search.html.twig#L62

UCPKN ticket: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/UCPKN-2923

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/card/pattern-card.html.twig
+++ b/templates/patterns/card/pattern-card.html.twig
@@ -43,10 +43,10 @@
   'title_tag': title_tag,
   'subtitle': (subtitle is not empty) ? {
     'content': subtitle,
-    'classes': 'mb-2',
   } : {},
   'text': (text is not empty) ? {
     'content': text,
+    'classes': 'mt-2',
     'tag': 'div',
   } : {},
   'content': _content,
--- a/templates/patterns/card/pattern-card.html.twig
+++ b/templates/patterns/card/pattern-card.html.twig
@@ -20,7 +20,7 @@
     {{ content }}
   {% endset %}
   {% if _rendered_content|trim is not empty %}
-    <div class="mt-2-5">{{ _rendered_content }}</div>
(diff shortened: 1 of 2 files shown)


templates/patterns/card/pattern-card.html.twig
```diff
@@ -20,7 +20,7 @@
     {{ content }}
   {% endset %}
   {% if _rendered_content|trim is not empty %}
-    <div>{{ _rendered_content }}</div>
+    <div class="mt-2-5">{{ _rendered_content }}</div>
```
drishu: so what if I send only content now, without meta? we get the same situation as reported but in a different combination of fields set. We need to resolve this without introducing new issues.

drishu (changes requested):
also the second point was not addressed

templates/patterns/card/pattern-card.html.twig
tibi2303: now it should not be a problem because the margin is top... also the margin between elements are eaten by themselves and only the bigger one is kept ex mb-3 vs mb-2-5 - the margin will be 16px (mb-3)

drishu (approved):
I see no changes where needed in the tests, but since its just some utility I think its ok

AaronGilMartinez (approved):

PR #427, opened 2024-08-29
OEL-3325: Responsive inpage navigation.

tibi2303:
Jira issue(s):

### DIFF
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -162,7 +162,7 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
  * Implements hook_preprocess_HOOK() for inpage navigation pattern.
  */
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
-  $variables['inpage_navigation_dropdown_id'] = $variables['attributes']['id'] . '-dropdown' ?? Html::getUniqueId('bcl-inpage-navigation') . '-dropdown';
+  $variables['inpage_navigation_dropdown_id'] = Html::getUniqueId(($variables['attributes']['id'] ?? 'bcl-inpage-navigation') . '-dropdown');
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
 }
 
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -162,6 +162,7 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
  * Implements hook_preprocess_HOOK() for inpage navigation pattern.
  */
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
+  $variables['inpage_navigation_dropdown_id'] = $variables['attributes']['id'] . '-dropdown' ?? Html::getUniqueId('bcl-inpage-navigation') . '-dropdown';
   $variables['inpage_navigation_id'] = $variables['attributes']['id'] ?? Html::getUniqueId('bcl-inpage-navigation');
 }
 
--- a/resources/sass/overrides/_toolbar.scss
+++ b/resources/sass/overrides/_toolbar.scss
@@ -9,6 +9,7 @@
 }
 .toolbar-fixed {
   .bcl-sidebar {
+    top: 38px;
     .bcl-inpage-navigation {
       top: 3rem;
     }
--- a/resources/sass/overrides/_toolbar.scss
+++ b/resources/sass/overrides/_toolbar.scss
@@ -7,6 +7,13 @@
     z-index: 5000;
   }
 }
+.toolbar-fixed {
+  .bcl-sidebar {
+    .bcl-inpage-navigation {
+      top: 3rem;
+    }
+  }
+}
 
 // Offcanvas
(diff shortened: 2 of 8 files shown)


includes/pattern.inc
```diff
@@ -162,6 +162,7 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
  * Implements hook_preprocess_HOOK() for inpage navigation pattern.
  */
 function oe_bootstrap_theme_preprocess_pattern_inpage_navigation(&$variables): void {
+  $variables['inpage_navigation_dropdown_id'] = $variables['attributes']['id'] . '-dropdown' ?? Html::getUniqueId('bcl-inpage-navigation') . '-dropdown';
```
brummbar: We always need to run everything through getUniqueId().
```suggestion
  $variables['inpage_navigation_dropdown_id'] = Html::getUniqueId(($variables['attributes']['id'] ?? 'bcl-inpage-navigation') . '-dropdown');
```

brummbar: I checked locally and indeed this pops up 
image

brummbar (changes requested):
It looks good, but two remarks:
* miss tests for the ID at least, possibly icon, and the non-presence of the d-none class.
* the mockups have a bottom border for the nav on mobile viewport.
* let's add a css in this theme to account for when the administration bar is present;
* let's add css to account for when the globan is present.

includes/pattern.inc
tibi2303: done, all the points.

brummbar:
Turns out the globan is not sticky on top so there's no conflict, since we cannot position an inpage nav before the header.
The code for the toolbar doesn't work for me:
image

tibi2303:
can be reviewed again

resources/sass/overrides/_toolbar.scss
```diff
@@ -16,6 +24,13 @@
   --bs-offcanvas-zindex: 5070;
 }
 
+body:has(#globan) {
+  .bcl-inpage-navigation,
+  .bcl-sidebar {
+    top: 28px;
+  }
+}
+
```
drishu: this is not needed in the end

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
```diff
@@ -73,3 +73,36 @@ inpage_navigation_with_no_links:
       class: Drupal\Tests\oe_bootstrap_theme\PatternAssertion\InPageNavigationAssert
       expected:
         title: 'Page content'
+inpage_navigation_with_dropdown_id_and_id:
```
drishu, line 78: dont think this is actually needed, nice the things are not new pattern fields, there is no actual new scenario, just add check to the assertBaseElements and we're fine

tests/src/PatternAssertion/InPageNavigationAssert.php
```diff
@@ -23,6 +23,16 @@ protected function getAssertions(string $variant): array {
...
+      'id' => [
+        [$this, 'assertElementAttribute'],
+        '.bcl-inpage-navigation',
+        'id',
+      ],
+      'dropdown_id' => [
+        [$this, 'assertElementAttribute'],
+        '.dropdown-toggle',
+        'id',
+      ],
```
drishu, line 35: you can just do this in assertBaseElements

tests/src/PatternAssertion/InPageNavigationAssert.php
```diff
@@ -31,8 +41,15 @@ protected function getAssertions(string $variant): array {
...
     $crawler = new Crawler($html);
+
     $inpage_nav = $crawler->filter('ul.nav-pills');
     self::assertCount(1, $inpage_nav);
+
+    $svg_icon = $crawler->filter('h2.bcl-heading svg');
+    self::assertCount(1, $svg_icon);
+
+    $base_element = $crawler->filter('.bcl-inpage-navigation');
+    self::assertStringNotContainsString('d-none', $base_element->attr('class'));
```
drishu, line 52: think the negative asset is not really needed just because we removed this utlitity cass

drishu (dismissed):

resources/sass/overrides/_toolbar.scss
tibi2303: i talked with fra, globan can be sticky and we should keep.

tests/src/Kernel/fixtures/markup_rendering_patterns/inpage_navigation.yml
brummbar: That's correct for the base elements, but I would keep the scenario, just use custom values for both of them to see that the value passed is respected.

tests/src/PatternAssertion/InPageNavigationAssert.php
tibi2303: miss tests for the ID at least, possibly icon, and the non-presence of the d-none class.

Fra added this point in the first review..

tests/src/PatternAssertion/InPageNavigationAssert.php
brummbar: We actually can't, as the value changes and that method cannot access any values passed.

brummbar (approved):

Issue #419, opened 2024-08-12
Status messages styling broken with big pipe after 10.3 drupal core update


vever001:
This is a preliminary report (will update once/if we start working on this).

After updating to drupal core 10.3, if the website uses big_pipe, it seems status messages are no longer styled (bypassing status-messages.html.twig) and show up unstyled/raw.
This seems to come from https://www.drupal.org/project/drupal/issues/3379885 which already has some reports of similar cases. 

This may require implementing `Drupal.theme.message` as per https://www.drupal.org/project/drupal/issues/3456176#comment-15692107

drishu:
needs more info, like steps to reproduce including what is expected versus what is actually.

vever001:
This is very easy to reproduce:
 - Install clean drupal core 10.2.x (standard profile*), oe_bootstrap_theme + helper, set theme as default
 - Login as superadmin, add a new page, notice the status message is OK/styled
![image](https://github.com/user-attachments/assets/4f3b5595-8286-49f5-85cf-cc4ecfb373a1)
- Now update to 10.3.x, create or update a basic page, notice status messages are unstyled, broken
![image](https://github.com/user-attachments/assets/ba3c239b-2e01-4fbf-a7a4-31d9554157d7)


[3379885](https://www.drupal.org/project/drupal/issues/3379885) was introduced in 10.3.0
*Big pipe comes enabled by default with the standard profile

enriquelacoma:
The issue was fixed in https://www.drupal.org/project/drupal/issues/3379885

vever001:
> The issue was fixed in https://www.drupal.org/project/drupal/issues/3379885

No, this is still relevant. Please reopen the issue.
I just reproduced this now with the oe_bootstrap_theme docker install and using the standard profile.
As stated, https://www.drupal.org/project/drupal/issues/3379885 is the issue that introduced the regression, on 10.3.0, which happens when bigpipe is enabled. 

See also https://www.drupal.org/project/drupal/issues/3457518 and https://www.drupal.org/project/drupal/issues/3396318
For now, since #3396318 is still ongoing, the only fix is to override  `Drupal.theme.message`, like many other themes did.

vever001:
cc @enriquelacoma @drishu
Thank you

vever001:
Could we please reopen? I do not have permissions. This issue is still valid.
FWIW I left a review in the opened PR. We can improve it further.
[...]

PR #428, opened 2024-09-05
Status messages styling broken with big pipe after 10.3 drupal core update

ioanaa20:
Jira issue(s):
Worked on the issue created here:
https://github.com/openeuropa/oe_bootstrap_theme/issues/419

### DIFF
--- a/oe_bootstrap_theme.theme
+++ b/oe_bootstrap_theme.theme
@@ -61,6 +61,51 @@ function oe_bootstrap_theme_theme_suggestions_views_view_table_alter(array &$sug
  * Implements hook_preprocess().
  */
 function oe_bootstrap_theme_preprocess(&$variables) {
+  // Construct the full URL for the BCL icons.
+  $variables['bcl_icon_path'] = _oe_bootstrap_theme_get_icon_path();
+}
+
+/**
+ * Implements hook_preprocess_HOOK() for table--bootstrap--responsive.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_table__bootstrap__responsive(&$variables) {
+  _oe_bootstrap_theme_bootstrap_responsive_table($variables);
+}
+
+/**
+ * Implements hook_preprocess_HOOK() for views-view-table--bootstrap--responsive.html.twig.
+ */
+function oe_bootstrap_theme_preprocess_views_view_table__bootstrap__responsive(&$variables) {
+  _oe_bootstrap_theme_bootstrap_responsive_table($variables);
+}
+
+/**
+ * Function that applies responsiveness to Bootstrap tables.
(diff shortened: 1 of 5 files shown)


oe_bootstrap_theme.theme
```diff
@@ -61,6 +62,51 @@ function oe_bootstrap_theme_theme_suggestions_views_view_table_alter(array &$sug
...
+}
+
+/**
+ * Function to generate the BCL icon path.
+ *
+ * @return string
+ *   The full URL for the BCL icons.
+ */
+function _oe_bootstrap_theme_get_icon_path(): string {
+  $path = &drupal_static(__FUNCTION__);
```
vever001, line 106: Nice find here +1, this is called hundreads of times per page load, all to return the same result everytime, so statically caching this makes sense.

This does create some noise in the PR and could also belong to a dedicated issue if maintainers prefer that.

oe_bootstrap_theme.theme
```diff
@@ -78,7 +124,7 @@ function oe_bootstrap_theme_preprocess(&$variables) {
 
     // Check if the BCL icon path is set in the current theme.
     if (!empty($theme_info['openeuropa']['bootstrap_component_library']['icon_path'])) {
-      $bcl_icon_path = $theme_info['openeuropa']['bootstrap_component_library']['icon_path'];
+      $bcl_icon_path = Html::escape($theme_info['openeuropa']['bootstrap_component_library']['icon_path']);
```
vever001, line 127: I'm not sure we need this actually.
This value is coming from code (theme .info.yml file so not really user input).

In twig this is already escaped on output, bcl-icon.html.twig:33
In JS I think it is better to use document.createElement + element.settAttribute, instead of element.innerHtml and a string literal as it's done right now (`messageWrapper.innerHTML += svgIconHtml;`).

resources/js/messages.js
```diff
@@ -0,0 +1,111 @@
...
+
+      closeButton.addEventListener("click", () => {
+        messageWrapper.classList.add("hidden");
+      });
+      messageWrapper.appendChild(closeButton);
+    }
+
+    return messageWrapper;
+  };
+})(Drupal, drupalSettings);
```
vever001, line 111: I did not check the JS implementation in detail but I think we should consider copy/pasting the one from ui_suite_bootstrap instead which seems to cover a lot more things.
https://git.drupalcode.org/project/ui_suite_bootstrap/-/blob/5.2.x/assets/js/misc/message.js?ref_type=heads
Radix uses the same (copy pasted from ui_suite_bootstrap): https://git.drupalcode.org/project/radix/-/blob/6.0.x/js/message.js?ref_type=heads

PR #430, opened 2024-09-11
OEL-3343: Copyright overlay.

tibi2303:
Jira issue(s):


templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
```diff
@@ -0,0 +1,14 @@
...
+    trigger_label:
+      type: text
+      label: 'Trigger label'
+      description: 'The label of the modal trigger. Optional.'
+      preview: 'View more'
+    text:
+      type: render
+      label: 'Text'
+      description: 'Copyright overlay text.'
+      preview: '© Lorem ipsum amet <a href="#">Amirhossein Soltani</a> on <a href="#">Getty Images</a>'
```
drishu: Dont think using real names is a good idea

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+/**
+ * @file
+ * Copyright overlay pattern.
+ */
+#}
+{{ attach_library('oe_bootstrap_theme/copy_clipboard') }}
+
+{% set _trigger_label = trigger_label|default('View more'|t) %}
+
+{% set _classes = ['copyright-overlay', 'px-2', 'text-muted'] %}
```
drishu: dont like the use of the utility classes, those should be used in other contexts, I'd like to keep it to semantic classes, add whats needed in css

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+{% set _footer %}
+  {% include '@oe-bcl/button' with {
+    label: 'Copy Copyright'|t,
+    attributes:  create_attribute()
+      .setAttribute('data-copy-target', 'copyright-text'),
+  } only %}
+{% endset %}
+
+{% set _header %}
+  <h5 class="mb-0">{{ 'Copyright'|t }}</h5>
```
drishu: same, the mb-0 utility

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+  } only %}
+{% endset %}
+
+{% set _body %}
+  <div class="copyright-text">
+    {{ text }}
+  </div>
+{% endset %}
+
+<div{{ attributes.addClass(_classes) }}>
```
drishu:
```suggestion
<div{{ attributes.addClass('copyright-overlay') }}>
```

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+  <span class="preview">
+    {{ text }}
+  </span>
+  {{ pattern('link', {
+    label: _trigger_label,
+    path: '#',
+    attributes: create_attribute()
```
drishu: I recommend trigger_attributes field

resources/sass/components/_copyright_overlay.scss
```diff
@@ -0,0 +1,11 @@
+.copyright-overlay {
+  background-color: rgba(var(--bs-white-rgb),var(--bs-bg-opacity));
+  .preview {
+    display: inline-block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    max-width: 108px;
```
drishu: 1. in the figma design, more of the "Lorem ipsum" phrase is visible
2. also if I look at figma the spacing is bigger for the content in the modal

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,42 @@
...
+  Drupal.behaviors.copy_clipboard = {
+    attach: function (context) {
+      Array.prototype.forEach.call(document.querySelectorAll('[data-copy-target]'), function (element) {
+        var targetClass = element.getAttribute('data-copy-target');
+        var targetElement = document.querySelector('.' + targetClass);
+        if (targetElement) {
+          element.addEventListener('click', function () {
+            var copyText = targetElement.innerText || targetElement.value;
+
+            // Copy the text to the clipboard
```
drishu: either add a dot at the end or remove the comment

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,42 @@
...
+   * Attaches the copy-to-clipboard behavior to elements with the 'data-copy-target' attribute.
+   *
+   * @type {Drupal~behavior}
+   *
+   * @prop {Drupal~behaviorAttach} attach
+   *   Initializes copy-to-clipboard functionality for each element.
+   * @prop {Drupal~behaviorDetach} detach
+   *   Cleans up any event listeners or instances to avoid memory leaks.
+   */
+  Drupal.behaviors.copy_clipboard = {
```
drishu:
```suggestion
  Drupal.behaviors.copyClipboard = {
```

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
drishu: for me the title can also be a field and default it to this

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+      .setAttribute('data-bs-toggle', 'modal')
+      .setAttribute('data-bs-target', '#' ~ modal_id),
+  }) }}
+</div>
+{% include '@oe-bcl/modal' with {
+  'id': modal_id,
+  'header': _header,
+  'show_close_button': true,
```
drishu: This is a field of the modal pattern, not really used like this.
I am a bit conflicted, should we not use the pattern? in case there are fixes or improvements it would resolve also here, feel this is half the reason to use patterns.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,57 @@
...
+    clean_class: true,
+    attributes:  create_attribute()
+      .addClass(['btn-close'])
+      .setAttribute('data-bs-dismiss', 'modal')
+      .setAttribute('aria-label', 'Close'),
+  } only %}
+{% endset %}
+
+{% set _body %}
+  <div class="copyright-text">
```
drishu: cant remember how we named it in the card v2, but I feel it should be body or content and not "text" as it can be html/render array

drishu (changes requested):
<s>when I click the copy button, it gives no feedback, feels strange</s> its fixed

templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
tibi2303: changed them.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: ok, yes.. i was thinking to about the v2 approach.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: actually the whole header is no longer needed, i am using directly the modal pattern.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done.

resources/sass/components/_copyright_overlay.scss
tibi2303: it is now

resources/js/copy_clipboard.js
tibi2303: removed.

resources/js/copy_clipboard.js
tibi2303: done.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: added the title as a field too

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: i am using now the pattern.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done, changed it to content.

resources/sass/components/_copyright_overlay.scss
```diff
@@ -0,0 +1,20 @@
...
+  background-color: rgba(var(--bs-white-rgb),var(--bs-bg-opacity));
+  color: var(--bs-gray-700);
+  padding: 0 0.5rem;
+  display: inline-flex;
+  .preview {
+    display: inline-block;
+    overflow: hidden;
+    text-overflow: ellipsis;
+    white-space: nowrap;
+    max-width: 112px;
```
drishu: should this be rem so its a relative value to the device viewing?

tests/src/Kernel/fixtures/markup_rendering_patterns/copyright_overlay.yml
```diff
@@ -0,0 +1,54 @@
+copyright_overlay:
+  render:
+    '#type': pattern
+    '#id': copyright_overlay
+    '#fields':
+      text: 'Lorem Ipsum dolor sit amet'
+  assert:
+    contains:
+      'span.preview': 'Lorem Ipsum dolor sit amet'
+      'div.copyright-text': 'Lorem Ipsum dolor sit amet'
```
drishu: we dont have copyright-text anymore

tests/src/Kernel/fixtures/markup_rendering_patterns/copyright_overlay.yml
```diff
@@ -0,0 +1,54 @@
+copyright_overlay:
```
drishu, line 1: you need to add this file to the list in the MarkupRenderingTest so it actually gets used

drishu (changes requested):

templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
```diff
@@ -0,0 +1,24 @@
...
+      type: text
+      label: 'Modal title'
+      description: 'The title of the modal header. Optional.'
+      preview: 'Copyright'
+    content:
+      type: render
+      label: 'Content'
+      description: 'Copyright overlay content.'
+      preview: '© Lorem ipsum amet <a href="#">John Doe</a> on <a href="#">Doe Images</a>'
+  settings:
```
drishu, line 20:
```
    attributes:
      type: attributes
      label: Copyright overlay attributes
      description: Extra attributes for the copyright overlay
      allow_token: true
```

templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
```diff
@@ -0,0 +1,24 @@
...
+    trigger_attributes:
+      type: attributes
+      label: Trigger attributes
+      description: Extra attributes for the modal trigger
```
drishu, line 24: allow_token: true

resources/sass/components/_copyright_overlay.scss
drishu: the bottom spacing is not there yet

resources/sass/components/_copyright_overlay.scss
tibi2303: changed it to rem. indeed

tests/src/Kernel/fixtures/markup_rendering_patterns/copyright_overlay.yml
tibi2303: changed and fixed the tests.

tests/src/Kernel/fixtures/markup_rendering_patterns/copyright_overlay.yml
tibi2303: added.

templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
tibi2303: added.

templates/patterns/copyright_overlay/copyright_overlay.ui_patterns.yml
tibi2303: added

drishu (dismissed):

resources/sass/components/_copyright_overlay.scss
```diff
@@ -0,0 +1,20 @@
+.copyright-overlay {
```
brummbar, line 1: This needs to be a dedicated library to attach to the component alone.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,48 @@
...
+
+{% set _trigger_label = trigger_label|default('View more'|t) %}
+{% set _modal_title = modal_title|default('Copyright'|t) %}
+
+{% set _footer %}
+  {% include '@oe-bcl/button' with {
+    label: 'Copy Copyright'|t,
+    attributes:  create_attribute()
+      .setAttribute('data-bs-dismiss', 'modal')
+      .setAttribute('data-copy-target', 'copyright-content'),
```
brummbar:
```suggestion
    attributes:  create_attribute({
      'data-bs-dismiss': 'modal',
      'data-copy-target': 'copyright-content'
    })
```

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,48 @@
...
+      .setAttribute('data-bs-toggle', 'modal')
+      .setAttribute('data-bs-target', '#' ~ modal_id),
+  }) }}
+  {{ pattern('modal', {
+    'title': _modal_title,
+    'show_close_button': true,
+    'body': _content,
+    'footer': _footer,
+    'attributes': create_attribute()
+      .setAttribute('id', modal_id)
```
brummbar:
```suggestion
    'attributes': create_attribute({
      id: modal_id
    })
```

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+   * @type {Drupal~behavior}
+   *
+   * @prop {Drupal~behaviorAttach} attach
+   *   Initializes copy-to-clipboard functionality for each element.
+   * @prop {Drupal~behaviorDetach} detach
+   *   Cleans up any event listeners or instances to avoid memory leaks.
+   */
+  Drupal.behaviors.copyClipboard = {
+    attach: function (context) {
+      Array.prototype.forEach.call(document.querySelectorAll('[data-copy-target]'), function (element) {
```
brummbar:
```suggestion
      once('oebt-clipcopy', '[data-copy-target]', context).forEach(function (element) {
```

brummbar: And of course we need to add once library to the dependencies.

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+            navigator.clipboard.writeText(copyText);
+          });
+        }
+      });
+    },
+
+    detach: function (context, settings, trigger) {
+      if (trigger === 'unload') {
+        Array.prototype.forEach.call(document.querySelectorAll('[data-copy-target]'), function (element) {
+          element.removeEventListener('click');
```
brummbar: Here we remove all click events. To be more accurate, we should remove only our click event. In order to do so, you need to pass the same callback that you used in the addEventListener method. It's enough to name the function and put it in the current scope.

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+            var copyText = targetElement.innerText || targetElement.value;
+            navigator.clipboard.writeText(copyText);
+          });
+        }
+      });
+    },
+
+    detach: function (context, settings, trigger) {
+      if (trigger === 'unload') {
+        Array.prototype.forEach.call(document.querySelectorAll('[data-copy-target]'), function (element) {
```
brummbar: This needs to use context.

oe_bootstrap_theme.libraries.yml
```diff
@@ -21,3 +21,7 @@ drupal.progress:
 inpage_navigation:
   js:
     resources/js/inpage_navigation.js: {}
+
+copy_clipboard:
+  js:
+    resources/js/copy_clipboard.js: {}
```
brummbar, line 29: Need to add also core/drupal as based dependency, so JS scripts are sorted accordingly.

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,40 @@
...
+   * @prop {Drupal~behaviorAttach} attach
+   *   Initializes copy-to-clipboard functionality for each element.
+   * @prop {Drupal~behaviorDetach} detach
+   *   Cleans up any event listeners or instances to avoid memory leaks.
+   */
+  Drupal.behaviors.copyClipboard = {
+    attach: function (context) {
+      Array.prototype.forEach.call(document.querySelectorAll('[data-copy-target]'), function (element) {
+        var targetClass = element.getAttribute('data-copy-target');
+        var targetElement = document.querySelector('.' + targetClass);
```
brummbar: One big problem here: if there are two copyright elements with the same class, the first one in the page will be always copied. If we want to be similar to bootstrap, let's just use the whole selector for the query, and leave the choice to the user. Otherwise we need to make sure to use an ID. Maybe ID in this case is better, or users won't know and pass a class and encounter the same issue.

resources/sass/components/_copyright_overlay.scss
brummbar: This library will have the standalone js and the copy_clipboard one as dependency.

brummbar (changes requested):
We need to write a JS test to test the functionality alone. Use the pattern preview page for the test.

resources/sass/components/_copyright_overlay.scss
tibi2303: done, have a look, i hope i understood corectly

resources/js/copy_clipboard.js
tibi2303: done

resources/js/copy_clipboard.js
tibi2303: done.

oe_bootstrap_theme.libraries.yml
tibi2303: done.

resources/js/copy_clipboard.js
tibi2303: changed.. now it has the whole selector as attr

resources/js/copy_clipboard.js
tibi2303: added.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done.

oe_bootstrap_theme.libraries.yml
```diff
@@ -21,3 +21,17 @@ drupal.progress:
...
+  dependencies:
+    - core/drupal
+    - core/once
+
+pattern.copyright_overlay:
```
drishu, line 34: I dont see the need for a separate library, css and js can be defined under the same lib, or am I missing something?

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,36 @@
...
+      // Attach click event listener to elements with data-copy-target
+      once('oebt-clipcopy', '[data-copy-target]', context).forEach(function (element) {
+        element.addEventListener('click', handleClick);
+      });
+    },
+
+    detach: function (context, settings, trigger) {
+      if (trigger === 'unload') {
+        Array.prototype.forEach.call(context.querySelectorAll('[data-copy-target]'), function (element) {
+          element.removeEventListener('click', handleClick);
```
drishu: you are sending  handleClick but it is not defined in this visibility scope, it is only available in attach

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,36 @@
...
+ */
+(function (Drupal) {
+
+  Drupal.behaviors.copyClipboard = {
+    attach: function (context) {
+      // Use a scoped function to handle click events
+      const handleClick = function (event) {
+        const element = event.currentTarget;
+        const targetSelector = element.getAttribute('data-copy-target');
+        const targetElement = document.querySelector(targetSelector);
```
drishu: I've tested this with 2 copyright patterns on the page and it always prints the content of the first one.
I used this to print it twice:
```
{{ pattern('copyright_overlay', {
    trigger_label: 'Trigger label 1',
    modal_title: 'Modal title 1',
    content: 'Content 1',
  }) }}

  {{ pattern('copyright_overlay', {
    trigger_label: 'Trigger label 2',
    modal_title: 'Modal title 2',
    content: 'Content 2',
  }) }}
```
and if I do `console.log(targetSelector);` it prints ".copyright-content" no matter which I click

tests/src/FunctionalJavascript/CopyClipboardTest.php
```diff
@@ -0,0 +1,68 @@
+<?php
+
+declare(strict_types=1);
+
+namespace Drupal\Tests\oe_bootstrap_theme\FunctionalJavascript;
+
+use Drupal\FunctionalJavascriptTests\WebDriverTestBase;
+
+/**
+ * Tests the copy to clipboard functionality.
```
drishu:
```suggestion
 * Tests the copy to clipboard javascript behavior.
```

tests/src/FunctionalJavascript/CopyClipboardTest.php
```diff
@@ -0,0 +1,68 @@
...
+    'ui_patterns_library',
+  ];
+
+  /**
+   * Tests the copy to clipboard behavior in the copyright overlay.
+   */
+  public function testCopyToClipboard(): void {
+    $assert_session = $this->assertSession();
+
+    // Create and log in an admin user with full permissions.
```
drishu: comments we should add to add clarification where something is ambiguous or hard to understand.. where it brings extra value in short. These comments here and not needed.

tests/src/FunctionalJavascript/CopyClipboardTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+    // Open the modal by clicking the trigger.
+    $assert_session->elementExists('css', '.copyright-trigger')->click();
+    $assert_session->waitForElementVisible('css', '.modal.show');
+
+    // Mock the clipboard API.
+    $this->getSession()->executeScript('
+      navigator.clipboard = {
+        writeText: function(text) {
+          window.copiedText = text; // Store the copied text in a variable
```
drishu: I've tried `navigator.clipboard.readText()` but its not available so nice workaround 👍

tests/src/FunctionalJavascript/CopyClipboardTest.php
```diff
@@ -0,0 +1,68 @@
...
+
+    // Ensure the copy button exists and simulate a click.
+    $assert_session->elementExists('css', '[data-copy-target=".copyright-content"]')->click();
+
+    // Now get the text from the mocked clipboard.
+    $expectedText = '© Lorem ipsum amet John Doe on Doe Images';
+    $actualText = $this->getSession()->evaluateScript('return window.copiedText;');
+
+    // Assert the copied text matches the expected text.
+    $this->assertEquals($expectedText, $actualText);
```
drishu: The test is ok, but I wish we could add two of them on the page to cover the copy error

drishu (changes requested):

oe_bootstrap_theme.libraries.yml
brummbar: The copy_clipboard.js is generic, it can be used for other elements.

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,34 @@
...
+
+  Drupal.behaviors.copyClipboard = {
+    attach: function (context) {
+      once('oebt-clipcopy', '[data-copy-target]', context).forEach(function (element) {
+        element.addEventListener('click', handleClick);
+      });
+    },
+
+    detach: function (context, settings, trigger) {
+      if (trigger === 'unload') {
```
drishu: Why limit it to unload?

drishu (changes requested):

drishu (dismissed):

resources/js/copy_clipboard.js
drishu: I removed as I think we should detach in any case

tibi2303: Yes, thanks for the modifications.

tibi2303: Unload in this case used only when the page is navigated away or if an ajax call is done and the content is changed. But it's better without the trigger.

includes/pattern.inc
```diff
@@ -99,6 +99,14 @@ function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): voi
   $variables['accordion_id'] = Html::getUniqueId('bcl-accordion');
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_copyright_overlay(array &$variables): void {
+  $variables['modal_id'] = Html::getUniqueId('bcl-copyright-modal');
+  $variables['copy_selector'] = Html::getUniqueId('copy-selector');
```
brummbar: I would use copyright-content as ID.

resources/js/copy_clipboard.js
```diff
@@ -0,0 +1,32 @@
+/**
+ * @file
+ * Attaches behaviors to copy text to clipboard for elements with data-copy-target attribute.
+ */
+(function (Drupal) {
+
+  const handleClick = function (event) {
+    const element = event.currentTarget;
+    const targetSelector = element.getAttribute('data-copy-target');
+    const targetElement = document.querySelector(targetSelector);
```
brummbar, line 10: So we decided to go for the approach where is the user that needs to ensure to pass a selector that triggers one element only?

brummbar (changes requested):
Looks good, one remark and one question.

resources/js/copy_clipboard.js
tibi2303: yes, i think it's better because we let the user make extended selectors and it's up to them to provide the correct one.

includes/pattern.inc
tibi2303: changed.

brummbar (approved):

drishu (approved):

PR #438, opened 2024-10-01
OEL-3417: Added subtitle tag for featured media.

tibi2303:
Jira issue(s):


templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -33,7 +33,12 @@
     <div class="row">
       <div class="col-12 col-md-6 {{ text_alignment_class }}">
         {% if subtitle %}
-          <h5 class="text-secondary">{{ subtitle }}</h5>
+          {% if featured_media_subtitle_tag_h5 %}
+            {% set _subtitle_tag = 'h5' %}
+          {% else %}
+            {% set _subtitle_tag = subtitle_tag is not empty ? subtitle_tag : (title is not empty ? 'h3' : 'h2') %}
```
AaronGilMartinez: Lets use default filter for the subtitle_tag, seems a bit confusing a ternary operator inside another ternary.

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
```diff
@@ -263,3 +263,37 @@ featured_media_iframe_with_text_container_right_aligned_with_link_ratio_21x9:
...
+  render:
+    '#type': pattern
+    '#id': featured_media
+    '#fields':
+      title: ''
+      subtitle: 'Subtitle Here'
+  assertions:
+    count:
+      'h2.text-secondary': 1
+      'h2.bcl-heading': 0
```
AaronGilMartinez, line 299: Lets add cases for the bc settings on

AaronGilMartinez (changes requested):

tests/src/Kernel/BackwardCompatibility/FeaturedMediaPatternTest.php
```diff
@@ -0,0 +1,64 @@
...
+    $crawler = new Crawler($html);
+    $this->assertCount(1, $crawler->filter('h5.text-secondary'));
+    $this->assertEquals('Test Subtitle', $crawler->filter('h5.text-secondary')->text());
+
+    // Test with backward compatibility setting disabled (h3).
+    $this->setBackwardCompatibilitySetting('featured_media_subtitle_tag_h5', FALSE);
+    $html = $this->renderBuild($build);
+    $crawler = new Crawler($html);
+    $this->assertCount(0, $crawler->filter('h5.text-secondary'));
+    $this->assertCount(1, $crawler->filter('h3.text-secondary'));
```
AaronGilMartinez: Add an extra case for when the tag for the subtitle is set and bc setting too, one of them will take precedence.

AaronGilMartinez (changes requested):

AaronGilMartinez (dismissed):

includes/pattern.inc
```diff
@@ -156,6 +156,8 @@ function oe_bootstrap_theme_preprocess_pattern_featured_media(&$variables) {
     $variables['image'] = ImageValueObject::fromArray($variables['image']->getArray())
       ->toRenderArray();
   }
+
+  $variables['featured_media_subtitle_tag_h5'] = BackwardCompatibility::getSetting('featured_media_subtitle_tag_h5');
```
drishu: above is an isset condition related to images, linking this to that condition, dont think one should depends on the other

tests/src/Kernel/BackwardCompatibility/FeaturedMediaPatternTest.php
```diff
@@ -0,0 +1,66 @@
...
+namespace Drupal\Tests\oe_bootstrap_theme\Kernel\BackwardCompatibility;
+
+use Drupal\Tests\oe_bootstrap_theme\Kernel\AbstractKernelTestBase;
+use Drupal\Tests\oe_bootstrap_theme\Traits\BackwardCompatibilityTrait;
+use Symfony\Component\DomCrawler\Crawler;
+
+/**
+ * Tests backward compatibility for the featured media pattern.
+ */
+class FeaturedMediaPatternTest extends AbstractKernelTestBase {
```
drishu: I presume that this entire test was made for the backward compatibility setting to be toggled and check the rendering, please remove, that can be done easly from the markuprendering test

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
```diff
@@ -263,3 +263,37 @@ featured_media_iframe_with_text_container_right_aligned_with_link_ratio_21x9:
     contains:
       'div.col-12.col-md-6.order-md-2': 'Lorem Ipsum dolor sit amet.'
       'div.col-12.col-md-6.order-md-2 a': 'View all'
+featured_media_with_subtitle_and_subtitle_tag:
```
drishu, line 266: use
```
  backward_compatibility:
    featured_media_subtitle_tag_h5: true
```
and list all situations with and without the the thing enabled, so you have a permutation of possibilities, in all cases the override takes priority and then you have h2 h3 or h5 depending on the BC value and weather title is set or not

drishu (changes requested):

tests/src/Kernel/BackwardCompatibility/FeaturedMediaPatternTest.php
tibi2303: It was created because before the bc settings were not provided for patterns page... now i change the markup rendering method and it passing them, and moved them to featured media.

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
tibi2303: yeah

includes/pattern.inc
tibi2303: done, thanks.

AaronGilMartinez (dismissed):

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
```diff
@@ -263,3 +263,77 @@ featured_media_iframe_with_text_container_right_aligned_with_link_ratio_21x9:
...
+    '#type': pattern
+    '#id': featured_media
+    '#fields':
+      title: 'Main Title'
+      subtitle: 'Subtitle'
+  assertions:
+    count:
+      'h3.text-secondary': 1
+      'h2.bcl-heading': 1
+featured_media_with_subtitle_empty_title_and_subtitle_tag:
```
drishu:
```suggestion
featured_media_with_subtitle_empty_title:
```

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
```diff
@@ -263,3 +263,77 @@ featured_media_iframe_with_text_container_right_aligned_with_link_ratio_21x9:
...
+  render:
+    '#type': pattern
+    '#id': featured_media
+    '#fields':
+      title: 'Main Title'
+      subtitle: 'Subtitle'
+      subtitle_tag: 'h6'
+  assertions:
+    count:
+      'h5.text-secondary': 1
```
drishu: you set subtitle_tag: 'h6' and test for 'h5.text-secondary': 1, it should override

tests/src/Kernel/MarkupRenderingTest.php
```diff
@@ -162,6 +162,7 @@ public function markupRenderingProvider(): array {
         } while (isset($test_cases[$candidate_key]));
         $key = $candidate_key;
         $test_case['render'] = PatternTestDataMassager::massageData($pattern, $test_case['render']);
+        $test_case['bc_settings'] = $test_case['bc_settings'] ?? [];
```
drishu: worked without this line for me

drishu (changes requested):

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
tibi2303: it has bc on setting, it should not

tests/src/Kernel/MarkupRenderingTest.php
tibi2303: indeed, weird that for me didnt work at first when i tried...

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
tibi2303: changed.

tests/src/Kernel/fixtures/markup_rendering_patterns/featured_media.yml
drishu: "Add a BC setting to use h5 by default." it should be the default, not override everything

tibi2303: i think the description is wrong... the implementation is correct as it is.... the bc is to use it as the previous version was (with h5).. if i set it as you said if on a project they already set a subtitle tag and enable the bc, the tag would still be in there and not a h5 as it should

brummbar: The BC setting provides only the default. Since the subtitle_tag parameter was not there before, the BC setting is used to determine its default status.
If a user purposely passes any value to it, that's the value to be used, not the h5.

drishu (approved):

PR #444, opened 2024-10-16
OEL-3421: Added accessible toggle for modal and offcanvas.

tibi2303:
Jira issue(s):

### DIFF
--- a/oe_bootstrap_theme.libraries.yml
+++ b/oe_bootstrap_theme.libraries.yml
@@ -1,6 +1,6 @@
 bcl:
   js:
-    assets/js/oe-bcl-default.bundle.min.js: {}
+    assets/js/oe-bcl-default.bundle.js: {}
   css:
     theme:
       assets/css/oe-bcl-default.min.css: {}
@@ -21,3 +21,7 @@ drupal.progress:
 inpage_navigation:
   js:
     resources/js/inpage_navigation.js: {}
+
+accessible_toggle:
+  js:
+    resources/js/accessible_toggle.js: {}
--- a/oe_bootstrap_theme.libraries.yml
+++ b/oe_bootstrap_theme.libraries.yml
@@ -1,6 +1,7 @@
 bcl:
   js:
-    assets/js/oe-bcl-default.bundle.js: {}
+    assets/js/oe-bcl-default.bundle.min.js: {}
+    resources/js/accessible_toggle.js: {}
   css:
     theme:
       assets/css/oe-bcl-default.min.css: {}
@@ -21,7 +22,3 @@ drupal.progress:
 inpage_navigation:
   js:
     resources/js/inpage_navigation.js: {}
-
-accessible_toggle:
-  js:
-    resources/js/accessible_toggle.js: {}
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "1.6.0",
-    "@openeuropa/bcl-theme-default": "1.6.0",
+    "@openeuropa/bcl-builder": "0.3421.2024101161200",
+    "@openeuropa/bcl-theme-default": "0.3421.2024101161200",
     "bootstrap-ie11": "5.0.2",
     "chokidar-cli": "^3.0.0",
     "copyfiles": "2.4.1",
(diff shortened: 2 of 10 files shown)


package.json
```diff
@@ -15,8 +15,8 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "1.6.0",
-    "@openeuropa/bcl-theme-default": "1.6.0",
+    "@openeuropa/bcl-builder": "0.3421.2024101161200",
+    "@openeuropa/bcl-theme-default": "0.3421.2024101161200",
```
drishu: testing was requested, the ticket asked for "copyright page" but that's a mistake as its part of an epic and not merged yet, so we can use  the patterns page where there is a modal and 2 overlays available

drishu (changes requested):

oe_bootstrap_theme.libraries.yml
```diff
@@ -1,6 +1,6 @@
 bcl:
   js:
-    assets/js/oe-bcl-default.bundle.min.js: {}
```
drishu, line 3: why change from min?

resources/js/accessible_toggle.js
```diff
@@ -0,0 +1,24 @@
+/**
+ * @file
+ * Attaches behaviors for accessible modal and offcanvas triggers.
```
drishu: if we add "collapse" in the future then we need to update this comment, so lets make it more generic

resources/js/accessible_toggle.js
```diff
@@ -0,0 +1,24 @@
...
+ */
+(function (bootstrap, Drupal) {
+
+  /**
+   * Attaches the accessible toggle behavior to modal and offcanvas elements.
```
drishu: same, more generic

resources/js/accessible_toggle.js
```diff
@@ -0,0 +1,24 @@
...
+   *
+   * @type {Drupal~behavior}
+   *
+   * @prop {Drupal~behaviorAttach} attach
+   *   Initializes AccessibleToggle for modal and offcanvas elements.
+   * @prop {Drupal~behaviorDetach} detach
```
drishu: I dont see a detach

resources/js/accessible_toggle.js
```diff
@@ -0,0 +1,24 @@
...
+   *   (Optional) Detach logic can go here if needed in the future.
+   */
+  Drupal.behaviors.accessibleToggle = {
+    attach: function (context, settings) {
+      // Initialize AccessibleToggle for modal and offcanvas triggers.
```
drishu: remove this comment, does not bring additional clarifications

templates/patterns/modal/pattern-modal.html.twig
```diff
@@ -4,6 +4,7 @@
  * Modal pattern.
  */
 #}
+{{ attach_library('oe_bootstrap_theme/accessible_toggle') }}
```
drishu: what about language list modal for example, and there could be other modals that I miss, rather make this part of the bcl library so it gets loaded on every page and acts wherever needed.

drishu (changes requested):
tests are missing!

drishu (approved):

drishu (changes requested):
test missing

tests/src/FunctionalJavascript/AccessibleToggleTest.php
```diff
@@ -0,0 +1,74 @@
...
+    $this->assertEquals('false', $offcanvasTrigger->getAttribute('aria-expanded'));
+
+    $offcanvasTrigger->click();
+    $this->assertEquals('false', $modalTrigger->getAttribute('aria-expanded'));
+    $this->assertEquals('true', $offcanvasTrigger->getAttribute('aria-expanded'));
+
+    // Close offcanvas and check attribute updates.
+    $this->getSession()->getPage()->find('css', '.offcanvas-backdrop')->click();
+    $this->assertEquals('false', $modalTrigger->getAttribute('aria-expanded'));
+    $this->assertEquals('false', $offcanvasTrigger->getAttribute('aria-expanded'));
```
drishu: enhance assertAccessibleAttributes with a true or false flag, after use it with named argument so its clear what is expected, ex
```php
$this->assertAccessibleAttributes($modalTrigger, expanded: FALSE);
```

drishu (changes requested):

drishu (dismissed):

drishu (approved):

PR #452, opened 2024-12-09
OEL-3503: Support slim select V2

piotrsmykaj:
Jira issue(s):
- [OEL-3503](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-3503)

### DIFF
--- a/bcl-builder.config.js
+++ b/bcl-builder.config.js
@@ -26,8 +26,8 @@ module.exports = {
     },
     {
       from: ["node_modules/@openeuropa/bcl-theme-default/js/**"],
-      to: path.resolve(outputFolder, "assets/js"),
-      options: { up: true },
+      to: path.resolve(outputFolder, "assets"),
+      options: { up: 3 },
     },
     {
       from: ["node_modules/bootstrap-ie11/css/**"],
--- a/package.json
+++ b/package.json
@@ -15,8 +15,8 @@
     "production": "npm-run-all build:*"
   },
   "devDependencies": {
-    "@openeuropa/bcl-builder": "1.8.2",
-    "@openeuropa/bcl-theme-default": "1.8.2",
+    "@openeuropa/bcl-builder": "0.3501.202502052130",
+    "@openeuropa/bcl-theme-default": "0.3501.202502052130",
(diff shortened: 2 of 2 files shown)


bcl-builder.config.js
```diff
@@ -26,8 +26,8 @@ module.exports = {
     },
     {
       from: ["node_modules/@openeuropa/bcl-theme-default/js/**"],
-      to: path.resolve(outputFolder, "assets/js"),
-      options: { up: true },
+      to: path.resolve(outputFolder, "assets"),
```
drishu, line 29: I dont understand why we copy the js into assets instead of assets/js

drishu (changes requested):
we also need to use a dev release of the BCL branch to see it all in action

bcl-builder.config.js
piotrsmykaj: @drishu it's a way to copy the JS folder content **and subdirectories** into the assets/js. Today I rebased the branch because slim select 2 was included in BCL 1.8.0, released a day after my commit, so now you can pull the latest changes on the branch and check that the JS directory has JS files and slim-select-2 directory with slim select js file.

drishu (dismissed):

bcl-builder.config.js
drishu: The correct tag for now should be [0.3501.202412051700](https://www.npmjs.com/package/@openeuropa/bcl-bootstrap/v/0.3501.202412051700)

tibi2303: @drishu I made a correct prerelease now: 0.3501.202502031900

Please use this version. In 1.8.0 the slim-select-2 should not be available. I made a release 1.8.2 which corrects this.
@piotrsmykaj so please use that prerelease (0.3501.202502031900) also in this one there is the bug fix for the offcanvas.

piotrsmykaj: @drishu there is a newer version done by Tiberiu, with fixed style: 0.3501.202502052130

drishu (dismissed):

drishu (dismissed):

drishu (approved):

PR #453, opened 2024-12-11
OEL-3569: Changed dropdown trigger from span to anchors for accessibility.

tibi2303:


src/MenuPreprocess.php
```diff
@@ -38,11 +38,21 @@ public function preprocessMenu(array &$variables, string $hook): void {
         $sub_item = $this->menuLink($sub_item);
       }
 
+      $trigger = $this->menuLink([
```
drishu: this changes the tag from span to anchor, this could be a breaking change, isn't there a css solution we could use?

drishu (changes requested):
please create a BCL dev release so we can also see the other changes

src/MenuPreprocess.php
tibi2303: I changed the approach, I checked on local oesa that everything it's working now.

What it's happening here is that because of bclLink process of links, the dropdown triggers that does not have a path are rendered as a span.  I was thinking at first to sustain this behaviour on bcl but it was double triggering because it already had data-bs-toggle attribute. I think the current approach will for sure only target the nav items that have a below and does not have link in order to be rendered as a link and not span because of hardcoding # as path

drishu (dismissed):

drishu (approved):

PR #459, opened 2025-02-18
ERATRANS-1057: Added mega-menu pattern.

tibi2303:
Jira issue(s):

### DIFF
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -95,6 +95,13 @@ function oe_bootstrap_theme_preprocess_pattern_description_list(&$variables) {
 /**
  * Implements hook_preprocess_HOOK().
  */
+function oe_bootstrap_theme_preprocess_pattern_mega_menu(array &$variables): void {
+  $variables['mega_menu_id'] = Html::getUniqueId('bcl-mega-menu');
+}
+
+/**
+ * Implements hook_preprocess_HOOK().
+ */
 function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): void {
   $variables['accordion_id'] = Html::getUniqueId('bcl-accordion');
 }
--- a/templates/patterns/mega_menu/mega_menu.ui_patterns.yml
+++ b/templates/patterns/mega_menu/mega_menu.ui_patterns.yml
@@ -1,0 +1,75 @@
+mega_menu:
+  label: 'Mega Menu'
+  description: 'Unfolds when website visitors hover over, click, or tap on a high-level navigation bar option.'
+  fields:
+    button_label:
+      type: string
+      label: 'Mega menu trigger button label'
+      description: 'The label to show in the mega menu trigger button.'
+      preview: 'Mega menu toggle'
+    link:
+      type: boolean
+      label: 'Trigger as a link'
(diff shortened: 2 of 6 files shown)


includes/pattern.inc
```diff
@@ -92,6 +92,13 @@ function oe_bootstrap_theme_preprocess_pattern_description_list(&$variables) {
   }
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_mega_menu(array &$variables): void {
```
drishu, line 98: besides the standaone, we need to also display this inside a navbar, we can use a preview template for that and add some explanations and how tos

templates/patterns/mega_menu/mega_menu.ui_patterns.yml
```diff
@@ -0,0 +1,75 @@
+mega_menu:
+  label: 'Mega Menu'
+  description: 'Unfolds when website visitors hover over, click, or tap on a high-level navigation bar option.'
+  fields:
+    button_label:
+      type: string
+      label: 'Mega menu trigger button label'
+      description: 'The label to show in the mega menu trigger button.'
```
drishu, line 8:
```suggestion
      description: 'A trigger, usually a menu item with children.'
```

templates/patterns/mega_menu/mega_menu.ui_patterns.yml
```diff
@@ -0,0 +1,75 @@
...
+      preview: 'Mega menu toggle'
+    link:
+      type: boolean
+      label: 'Trigger as a link'
+      description: 'If enabled, the main trigger is rendered as a link. Otherwise, it is rendered as a button.'
+    back_button_label:
+      type: text
+      label: 'Back button label'
+      description: 'Text used for the “Back” button inside the menu.'
```
drishu, line 17:
```suggestion
      description: 'The back button appears only in smaller devices.'
```

tests/src/Kernel/fixtures/markup_rendering_patterns/mega_menu.yml
```diff
@@ -0,0 +1,70 @@
...
+      'button#bcl-mega-menu': 1
+      'div.mega-menu-container.dropdown-menu[aria-labelledby="bcl-mega-menu"]': 1
+      'button.back-button': 1
+      'div.navigation-items > div.dropdown': 2
+      'div.navigation-items > a': 4
+      'div.navigation-items > div.collapse': 1
+    equals:
+      'button.dropdown-toggle[data-bs-toggle="dropdown"]': 'Mega menu toggle'
+      'div.content-block': 'Content block description'
+      '.collapse': '<a href="#" class="mb-3 d-block">Example link</a><img src="https://inno-ecl.s3.amazonaws.com/media/examples/example-image.jpg" alt="image">'
```
drishu, line 70: we need to explore more variants here, what happens when the last col is empty, what if only first col has content, different content for the columns, and think if I missed any

drishu (changes requested):
since this is made in isolation of a menu (navbar) and not as a variant of a menu, we need to make sure they can work toghter

PR #460, opened 2025-02-24
EWPP-4655: Support D11

msnassar:


composer.json
```diff
@@ -7,23 +7,22 @@
...
         "php": ">=8.1",
-        "drupal/core": "^10",
-        "drupal/file_link": "^2.1",
-        "drupal/ui_patterns": "^1.4",
-        "drupal/ui_patterns_settings": "^2.0",
+        "cweagans/composer-patches": "^1.7",
+        "drupal/core": "^10.3 || ^11",
+        "drupal/file_link": "^2.2",
+        "drupal/ui_patterns": "^1.10",
+        "drupal/ui_patterns_settings": "^2.3",
```
kp77, line 14: Is it required to bump these versions? Just wondering, because if it's not necessary, we should leave it as is IMO.

msnassar: Since now the theme supports both D10 and D11, I would prefer to use versions that works for both versions. Besides it is cleaner it is also helpful when debugging composer with `composer why`

kp77: Thanks for the feedback, if the maintainers are fine with it, then ok for me.

kp77 (approved):

peterserfozo (approved):

drishu (approved):

PR #470, opened 2025-04-04
ELA-1304: Card with new date block display.

Aporie:
Jira issue(s):
- [ELA-1304](https://citnet.tech.ec.europa.eu/CITnet/jira/browse/ELA-1304)

In association with https://github.com/openeuropa/bootstrap-component-library/issues/656, this PR should produce a change in the date block of the card display:

Old:
image
image

New:
image
image

### DIFF
--- a/templates/patterns/card/pattern-card--variant-search.html.twig
+++ b/templates/patterns/card/pattern-card--variant-search.html.twig
@@ -48,6 +48,13 @@
   {% set _right_col_classes = 'col-md-9 col-lg-10' %}
 {% endif %}
 
+{% set _date = date|default({}) %}
+{% if _date is not empty %}
+  {% set _date = _date|merge({
+    icon_path: bcl_icon_path
+  })%}
+{% endif %}
+
 {% include '@oe-bcl/card' with {
   'horizontal': true,
   'horizontal_grid': {
@@ -77,7 +84,7 @@
     'path': image.src,
     'alt': image.alt,
   }) : {},
-  'date': date|default({}),
+  'date': _date,
   'content': _content,
 } only %}
 {% endapply %}
--- a/templates/patterns/card/pattern-card--variant-search.html.twig
+++ b/templates/patterns/card/pattern-card--variant-search.html.twig
@@ -44,8 +44,8 @@
 
 {# Backward compatibility setting: card uses grid classes. #}
 {% if bc_card_search_use_grid_classes %}
-  {% set _left_col_classes = 'col-md-3 col-lg-2 rounded mw-listing-img mb-md-0 mb-3' %}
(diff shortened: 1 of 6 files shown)


templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -48,6 +48,10 @@
   {% set _right_col_classes = 'col-md-9 col-lg-10' %}
 {% endif %}
 
+{% set _date = date|merge({
+  icon_path: bcl_icon_path
```
Aporie: Not sure it's the best approach.

templates/patterns/date_block/pattern-date-block.html.twig
```diff
@@ -11,4 +11,5 @@
   'year': date|date('Y'),
   'date_time': date|date('Y-m-d'),
   'attributes': attributes,
+  'icon_path': bcl_icon_path
```
Aporie, line 14: Not sure if we need that, or if it has to be applied elsewhere. I think the bcl-card component, will now expect a path to icons by default.

julien-:
Why not using card_v2 and only define the template of fields of type date ?

Aporie:
Card V2 does not use BCL component and seem to provide a single pattern file.

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -48,6 +48,13 @@
   {% set _right_col_classes = 'col-md-9 col-lg-10' %}
 {% endif %}
 
+{% set _date = date|default({}) %}
```
Aporie, line 51: Not very elegant and not sure it's the best approach.

templates/patterns/card/pattern-card--variant-search.html.twig
```diff
@@ -48,6 +48,13 @@
...
+{% if _date is not empty %}
+  {% set _date = _date|merge({
+    icon_path: bcl_icon_path
+  })%}
+{% endif %}
+
```
piotrsmykaj:
```suggestion
{% set _date = date|default({})|merge({ icon_path: bcl_icon_path }) %}

```
I would suggest with the simpler solution and add the icon_path param in any case.

Aporie: Yep, more elegant. I'll do.
[EDIT] now I remember. Problem doing that, is that icon_path will always be passed, making _date never empty, breaking other subtemplates ...

piotrsmykaj: Ok it makes sense, let's keep your code.

piotrsmykaj (dismissed):

tibi2303 (approved):

PR #484, opened 2025-05-20
OEL-3731: Make sure landmarks are unique.

vengador:
Jira issue(s):
- [OEL-3731](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-3731)

Fixed:
- Add aria-label attributes to main content and header.
- Add aria-label attribute to navigation bar (with letters and numbers) used in Glossary.

### DIFF
--- a/includes/pattern.inc
+++ b/includes/pattern.inc
@@ -227,10 +227,9 @@ function oe_bootstrap_theme_preprocess_pattern_listing__preview(array &$variable
 
 /**
  * Implements hook_preprocess_HOOK() for the pagination pattern.
- *
- * Adds default values for custom icons.
  */
 function oe_bootstrap_theme_preprocess_pattern_pagination(array &$variables): void {
+  // Adds default values for custom icons.
   foreach (['first', 'prev', 'next', 'last'] as $key) {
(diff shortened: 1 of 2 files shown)


includes/pattern.inc
```diff
@@ -240,6 +239,13 @@ function oe_bootstrap_theme_preprocess_pattern_pagination(array &$variables): vo
       'size' => 'fluid',
     ];
   }
+
+  // Add aria label for Glossary pagination.
+  if ($variables['variant'] !== 'glossary') {
+    return;
+  }
+
+  $variables['attributes']['aria-label'] = t('Glossary links');
```
AaronGilMartinez, line 248: I see that the pattern defines some aria attributes in the template, I'm wondering if there is an advantage of doing this here rather the template.

AaronGilMartinez (changes requested):

includes/pattern.inc
vengador: I personally consider that template should include those common attributes between different variants. This attribute is specific for glossary variant of that template so instead adding it in the template, I think it is better to include it in preprocess.

AaronGilMartinez (approved):

PR #502, opened 2025-07-11
OEL-3982: Offcanvas sticky-top on Desktop.

julien-:


allternativ9:
Hi Julien,

Many thanks for your proposal and contribution.
Following our team review, we have concluded that this feature does not meet the usability and accessibility requirements.
When the list of filters is long (as is the case on many projects), users are not able to see or access all available filters.

For these reasons,  we will not proceed with this PR for now. 
Many thanks for your understanding.
Angela.

PR #504, opened 2025-08-06
OEL-4016: Mega menu

donquixote:
Jira issue(s):
- [OEL-4016](https://webgate.ec.europa.eu/CITnet/jira/browse/OEL-4016)

### DIFF
--- a/modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.post_update.php
+++ b/modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.post_update.php
@@ -33,7 +33,7 @@ function oe_bootstrap_theme_helper_post_update_00002(): string {
   $report = 'The new mega menu block was placed in the navigation region for oe_bootstrap_theme.';
 
   $old_navigation_block = Block::load('oe_bootstrap_theme_mainnavigation');
-  if (!$old_navigation_block || !$old_navigation_block->getTheme() !== 'oe_bootstrap_theme') {
+  if (!$old_navigation_block || $old_navigation_block->getTheme() !== 'oe_bootstrap_theme') {
     return $report . "\nThe old navigation block was not found.";
   }
   if (!$old_navigation_block->status()) {
--- a/modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.post_update.php
+++ b/modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.post_update.php
@@ -7,6 +7,9 @@
 
 declare(strict_types=1);
 
+use Drupal\block\Entity\Block;
+use Drupal\oe_bootstrap_theme\ConfigImporter;
+
 /**
  * Uninstall OE Bootstrap Theme Paragraphs module.
  */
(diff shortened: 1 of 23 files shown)


modules/oe_bootstrap_theme_helper/oe_bootstrap_theme_helper.post_update.php
```diff
@@ -16,3 +19,28 @@ function oe_bootstrap_theme_helper_post_update_00001(array &$sandbox): void {
...
+  if (!\Drupal::moduleHandler()->moduleExists('block')) {
+    return 'No blocks can be placed, because the block module is not installed.';
+  }
+
+  ConfigImporter::importSingle('module', 'oe_bootstrap_theme_helper', '/config/post_updates/00002_megamenu', 'block.block.oe_bootstrap_theme_oelmegamenu');
+
+  $report = 'The new mega menu block was placed in the navigation region for oe_bootstrap_theme.';
+
+  $old_navigation_block = Block::load('oe_bootstrap_theme_mainnavigation');
+  if (!$old_navigation_block || !$old_navigation_block->getTheme() !== 'oe_bootstrap_theme') {
```
piotrsmykaj:
```suggestion
  if (!$old_navigation_block || $old_navigation_block->getTheme() !== 'oe_bootstrap_theme') {
```

modules/oe_bootstrap_theme_helper/src/Plugin/Block/MegaMenuBlock.php
```diff
@@ -0,0 +1,337 @@
...
+        }
+        $bcl_item['content_link'] = [
+          'label' => $this->t('Discover more'),
+          'path' => $more_url,
+        ];
+      }
+    }
+    elseif (!$is_nolink) {
+      $bcl_item['see_all'] = [
+        'label' => $this->t('More of @title', ['@title' => mb_lcfirst($item['title'])]),
```
piotrsmykaj: It's only available starting with PHP 8.3 and Drupal has its equivalent Unicode::lcfirst($string).

modules/oe_bootstrap_theme_helper/src/Plugin/Block/MegaMenuBlock.php
```diff
@@ -0,0 +1,337 @@
...
+   */
+  protected function getSystemMenuBlock(): ?SystemMenuBlock {
+    $menu_name = $this->configuration['menu'] ?? NULL;
+    if (!$menu_name) {
+      return NULL;
+    }
+    $system_menu_block = $this->blockManager->createInstance(
+      'system_menu_block:' . $menu_name,
+      [
+        'depth' => 3,
```
piotrsmykaj, line 128: Is it ok to hardcode a depth? 
How does Site Builder know about it? 
It may be confusing while a website has more than 3 levels in a menu structure.
Can the "depth" be added to the block.settings with 3 as a max level if it's not allowed to have more?

donquixote: The current design and front-end implementation only supports 3 levels (one horizontal and then two submenus inside the megamenu panel). Any deeper links are hidden.

modules/oe_bootstrap_theme_helper/src/Plugin/Block/MegaMenuBlock.php
donquixote: Good point!
`mb_lcfirst()` is available in PHP >= 8.4. Below that we would use a polyfill from symfony.
The polyfill is slightly more "advanced" than the Drupal core method, in that it determines the encoding first.
But Drupal core uses the Unicode::*() methods, so we should stick with the program.

modules/oe_bootstrap_theme_helper/src/Plugin/Block/MegaMenuBlock.php
piotrsmykaj: It's fine that it's consistent with the design and frontend implementation. As it's hardcoded, I would like to inform the site builder about the level limit (some websites may have more than 3 level menu), ex. on the block edit page. What do you think?

piotrsmykaj (approved):

PR #512, opened 2025-09-30
OEL-4178: Test with drupal/core:~11.2.0.

donquixote:
### Known issue

We have one deprecation notice in phpunit, in TwigExtensionTest::testElementChildrenFilter, for calling ->render($build, TRUE). It occurs if we run phpunit with `--fail-on-notice`.

> 1) /var/www/html/build/core/lib/Drupal/Core/Render/Renderer.php:223
> Drupal\Core\Render\Renderer::render with $is_root_call is deprecated in drupal:11.2.0 and is removed from drupal:12.0.0. Use Drupal\Core\Render\Renderer::renderRoot() instead.  See https://www.drupal.org/
> node/3497318.
> 
> Triggered by:
> 
> * Drupal\Tests\oe_bootstrap_theme_helper\Kernel\TwigExtensionTest::testElementChildrenFilter (8 times)
>  /var/www/html/modules/oe_bootstrap_theme_helper/tests/src/Kernel/TwigExtensionTest.php:242
> 

Here is the code snippet:

```php
        $render_context = new RenderContext();
        $output = $renderer->executeInRenderContext($render_context, function () use ($renderer, $build) {
          return $renderer->render($build, TRUE);
        });

        $bubbled_metadata = [];
        $render_context->pop()->applyTo($bubbled_metadata);
```
As we can see, we need the `$render_context` afterwards.

I did not find a suitable replacement for this code.
Simply replacing `->render(*, TRUE)` with `->renderRoot(*)` results in test failure.
When replacing the entire `->executeInRenderContext()` call, we don't have access to the `$render_context` anymore.

Besides, we still want this to work in 10.x.

This is only called in a test, and I don't think it is a problem.

### DIFF
--- a/.drone.yml
+++ b/.drone.yml
@@ -112,3 +112,5 @@ matrix:
       PHP_VERSION: 8.3
     - CORE_VERSION: 11.1.0
       PHP_VERSION: 8.3
+    - CORE_VERSION: 11.2.0
+      PHP_VERSION: 8.3
--- a/tests/src/Functional/ThemeSettingsTest.php
+++ b/tests/src/Functional/ThemeSettingsTest.php
@@ -5,12 +5,15 @@ declare(strict_types=1);
 namespace Drupal\Tests\oe_bootstrap_theme\Functional;
 
 use Drupal\Tests\BrowserTestBase;
+use Drupal\Tests\oe_bootstrap_theme\Traits\GetSelectOptionsTrait;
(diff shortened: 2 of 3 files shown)


.drone.yml
piotrsmykaj, line 1: I suggest removing the not-supported D10.3

piotrsmykaj: as discussed on Teams, we keep it for now.

piotrsmykaj (approved):

PR #520, opened 2025-10-23
OEL-3342: Copyright for images.

tibi2303:
We add a "copyright" or "image_copyright" parameter to multiple components.
The copyright text will be shown below or near the image.


tibi2303:
All the sub prs have been merged to this epic. It can be overall reviewed.

includes/pattern.inc
```diff
@@ -99,6 +99,14 @@ function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): voi
...
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_copyright_overlay(array &$variables): void {
+  $variables['modal_id'] = Html::getUniqueId('bcl-copyright-modal');
+  $variables['copy_selector'] = Html::getUniqueId('copyright-content');
+}
+
```
donquixote: We have to replace this preprocess when we go for ui_patterns v2.
For now perhaps we cannot avoid it.

includes/pattern.inc
```diff
@@ -99,6 +99,14 @@ function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): voi
   $variables['accordion_id'] = Html::getUniqueId('bcl-accordion');
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_copyright_overlay(array &$variables): void {
+  $variables['modal_id'] = Html::getUniqueId('bcl-copyright-modal');
+  $variables['copy_selector'] = Html::getUniqueId('copyright-content');
```
donquixote: One thing I wonder is how reliably unique are those ids if cache is involved.

recipes/media_av_portal_photo_copyright/config/field.field.media.av_portal_photo.field_media_copyright.yml
donquixote, line 1: Is oe_bootstrap_theme the right place for this?
What were the arguments/reasons to put it here?

templates/patterns/banner/pattern-banner.html.twig
```diff
@@ -19,16 +19,33 @@
...
+    title_tag: title_tag,
+    description: description,
+    link: call_to_action,
+    image: variant == 'image' or variant == 'image_shade' ? image : '' ,
+    content_classes: content_classes,
+    attributes: _banner_attributes,
+  } only %}
+{% endset %}
+{% if copyright is not empty %}
+  {% set _copyright = copyright.content is defined ? copyright : { content: copyright } %}
```
donquixote: Is this a case we have to / want to support, where copyright is already an object?
The schema says it should be a string. (or maybe just markup)

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,41 @@
...
+{{ attach_library('oe_bootstrap_theme/pattern.copyright_overlay') }}
+
+{% set _content %}
+  <div{{ create_attribute({
+    'class': 'copyright-content',
+    'id': copy_selector
+  }) }}>
+    {{ content }}
+  </div>
+{% endset %}
```
donquixote: We probably want whitespace removal here and elsewhere, given that we no longer use spaceless filter.


```suggestion
{% set _content -%}
  <div{{ create_attribute({
    'class': 'copyright-content',
    'id': copy_selector
  }) }}>
    {{- content -}}
  </div>
{%- endset %}
```

and/or
```suggestion
{% set _content -%}
  <div{{ create_attribute({
    'class': 'copyright-content',
    'id': copy_selector
  }) }}>{{ content }}</div>
{%- endset %}
```

donquixote (commented):
Some intermediate feedback. More later.

templates/patterns/banner/pattern-banner.html.twig
tibi2303: yes, indeed.. i think it's an overkill, i will change

includes/pattern.inc
tibi2303: I’ve made the IDs derive from attributes id. Defaults stay the same when no id is set.

tibi2303: I think it's good enough, if you think of something else let me know.

includes/pattern.inc
tibi2303: indeed, we can change further down the line

recipes/media_av_portal_photo_copyright/config/field.field.media.av_portal_photo.field_media_copyright.yml
tibi2303: it's related to the recipe.. i am not yet decided if you should move the recipe and config to whitelabel or keep it here... they are also related to the templates so that's why i am thinking oe_bt is a good place to have this.

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
tibi2303: done.

includes/pattern.inc
donquixote: The duplication would happen if Drupal puts two elements on the page that use the same cache id. But I don't know a scenario where that happens.

recipes/media_av_portal_photo_copyright/config/field.field.media.av_portal_photo.field_media_copyright.yml
donquixote: "related to the templates" - in general oebt has the component templates which can be included from content templates. So in that sense all the templates in oe_whitelabel are also related to oebt templates, because they include them..

includes/pattern.inc
```diff
@@ -99,6 +99,19 @@ function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): voi
...
 }
 
+/**
+ * Implements hook_preprocess_HOOK().
+ */
+function oe_bootstrap_theme_preprocess_pattern_copyright_overlay(array &$variables): void {
+  $base_id = NULL;
+  if (isset($variables['attributes']['id']) && $variables['attributes']['id'] !== '') {
+    $base_id = $variables['attributes']['id'];
+  }
```
donquixote: Is the `''` case relevant? (I guess so..)
How about this?
(it also replaces "0" because that is false-y, but I think that is ok.)

```suggestion
  $base_id = ($variables['attributes']['id'] ?? '') ?: 'bcl-copyright';
```
```php
```
Or if we want to keep the if/else:
```suggestion
  $base_id = 'bcl-copyright';
  if (isset($variables['attributes']['id']) && $variables['attributes']['id'] !== '') {
    $base_id = $variables['attributes']['id'];
  }
```
As a result the variable won't be NULL.

includes/pattern.inc
```diff
@@ -99,6 +99,19 @@ function oe_bootstrap_theme_preprocess_pattern_accordion(array &$variables): voi
...
+
+  $variables['modal_id'] = Html::getUniqueId(($base_id ? $base_id . '-modal' : 'bcl-copyright-modal'));
+  $variables['copy_selector'] = Html::getUniqueId(($base_id ? $base_id . '-content' : 'copyright-content'));
```
donquixote: Strange: In one case we have "bcl-" prefix, in the other we don't.
With the transformation above, and if we use "bcl-" prefix in both cases, we can now simplify:
```suggestion
  $variables['copy_selector'] = Html::getUniqueId(($base_id . '-content'));
```

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,41 @@
+{#
+/**
+ * @file
+ * Copyright overlay pattern.
+ */
+#}
+{{ attach_library('oe_bootstrap_theme/pattern.copyright_overlay') }}
+
+{% set _content -%}
```
donquixote:
```suggestion
{{- attach_library('oe_bootstrap_theme/pattern.copyright_overlay') }}

{%- set _content -%}
```
I see two options:
- We determine for each template and each position whether we need to remove whitespace or not.
- We always remove non-intentional whitespace.

donquixote: (in this case the whitespace would probably be ok because the next html tag is `<div>`.
On the other hand, that `<div>` is a modal and might be placed between inline elements. So might not be so clear.

templates/patterns/copyright_overlay/pattern-copyright-overlay--preview.html.twig
```diff
@@ -0,0 +1,14 @@
...
+ */
+#}
+
+{{ pattern('copyright_overlay', {
+  content: content,
+}) }}
+
+{{ pattern('copyright_overlay', {
+  content: '© Second copyright element. Suspendisse vel mauris vitae ipsum blandit condimentum ut eget quam.',
+}) }}
```
donquixote:
```suggestion
{{- pattern('copyright_overlay', {
  content: content,
}) }}

{{- pattern('copyright_overlay', {
  content: '© Second copyright element. Suspendisse vel mauris vitae ipsum blandit condimentum ut eget quam.',
}) }}
```

includes/pattern.inc
donquixote: Where is this id used?
I see it is added to a specific element, but why?

templates/patterns/copyright_overlay/pattern-copyright-overlay.html.twig
```diff
@@ -0,0 +1,41 @@
...
+  }) }}
+
+  {{- pattern('modal', {
+    'title': modal_title|default('Copyright'|t),
+    'show_close_button': true,
+    'body': _content,
+    'footer': _footer,
+    'attributes': create_attribute({
+      id: modal_id
+    })
```
donquixote: I notice that the modal shows up at the top of the viewport.
I think it would be nicer to have it centered.
What do you think? Do we have specs for this?

templates/patterns/carousel/pattern-carousel.html.twig
donquixote, line 1: We should add a copyright in the carousel preview data.

donquixote, line 1: Btw currently it is the same image each slide.
We can change it by appending `?seed=1234` to the url (or other seeds).
And maybe make it 3 slides instead of 2, so we can do some variation.
(you need to clear cache before you see the change)

donquixote, line 1: In the carousel pattern preview I cannot click the copyright in a slide, because of z-index, it is hidden behind the carousel buttons.

templates/patterns/gallery/pattern-gallery.html.twig
donquixote, line 1: Lets put a copyright in the preview data.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
```diff
@@ -137,6 +137,34 @@ private static function massageGalleryPattern(array $data): array {
...
+
+    if (isset($data['#fields']['copyright']['content']['#markup'])) {
+      $data['#fields']['copyright']['content'] = Markup::create($data['#fields']['copyright']['content']['#markup']);
+    }
+    elseif (isset($data['#fields']['copyright']['#markup'])) {
+      $data['#fields']['copyright'] = Markup::create($data['#fields']['copyright']['#markup']);
+    }
+
+    return $data;
+  }
```
donquixote: Do we need all of this massaging?
Can't we standardize the input structure for whether we need a 'content' key?
And then, the following just works in Drupal twig, or not? `{{ { '#markup' => 'Hello' } }}`.

tests/src/Kernel/fixtures/markup_rendering_patterns/media_container.yml
```diff
@@ -0,0 +1,79 @@
...
+    '#fields':
+      media:
+        '#markup': '<video controls><source src="movie.mp4" type="video/mp4"></video>'
+      description:
+        '#markup': '<p>Video description.</p>'
+      copyright:
+        content:
+          '#markup': '© 2024 Example Organisation'
+        trigger_label: 'Licence'
+        modal_title: 'Usage rights'
```
donquixote: Interesting, now I see why we have the sub-key for 'copyright'/'content'.
And I also see why you would support an alternative simpler structure, so that people can just pass `copyright: 'Abc'`.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
donquixote: I think/remember the main reason for this massaging is to allow for values that are not easy to express in yaml, is that correct?
So in the real world we would see a Markup object, but in yaml we can only put arrays, so `['#markup' => 'Abc']` becomes `Markup::create('Abc')`.

As an alternative, we could use yaml tagged value like this:
```yml
      copyright:
        content: !Markup '© 2024 Example Organisation'
```

donquixote: The transformation would look like this:
(We can drop the pattern-specific method)
```php
use Symfony\Component\Yaml\Tag\TaggedValue;
[...]
  /**
   * Massages test data for a pattern.
   *
   * @param string $patternName
   *   The pattern name.
   * @param array $data
   *   The data structure.
   *
   * @return array
   *   The massaged data structure.
   */
  public static function massageData(string $patternName, array $data): array {
    $callable = [
      self::class,
      'massage' . self::snakeToCamel($patternName) . 'Pattern',
    ];

    if (is_callable($callable)) {
      $data = call_user_func($callable, $data);
    }

    $data = self::massageDataRecursive($bkp = $data);
    
    return $data;
  }

  /**
   * Applies data massaging recursively.
   *
   * @param array $data
   *   The original data from the yaml file.
   *
   * @return array
   *   The massaged data structure.
   */
  protected static function massageDataRecursive(array $data): array {
    $result = [];
    foreach ($data as $key => $value) {
      if ($value instanceof TaggedValue) {
        $value = self::resolveTaggedValue($value->getTag(), $value->getValue());
      }
      elseif (is_array($value)) {
        $value = self::massageDataRecursive($value);
      }
      $result[$key] = $value;
    }
    return $result;
  }

  /**
   * Resolves a single tagged value.
   *
   * @param string $tag
   *   The tag name.
   * @param string $value
   *   The tagged value.
   *
   * @return mixed
   *   The resolved value to use as pattern input.
   */
  protected static function resolveTaggedValue(string $tag, string $value): mixed {
    return match ($tag) {
      'Markup' => Markup::create($value),
    };
  }
```
We could support more object types (ImageValueObject), but that would be out of scope for now.

tests/src/Kernel/fixtures/markup_rendering_patterns/banner.yml
```diff
@@ -198,6 +198,27 @@ primary_page_hero_banner_aligned_left_without_call_to_action_full_width:
...
+      title: 'EU Budget for the future'
+      description: 'Innovation, economy, environment and geopolitics'
+      copyright: '© 2024 Example Organisation'
+  assertions:
+    count:
+      'div.media-container': 1
+      'div.media-container .copyright-overlay': 1
+      'div.media-container div[data-drupal-selector=banner]': 1
+    contains:
+      '.copyright-overlay span.copyright-preview': '© 2024 Example Organisation'
```
donquixote: Does a copyright make sense in a banner with no image?

templates/patterns/media_container/pattern-media-container.html.twig
```diff
@@ -0,0 +1,40 @@
...
+  - media: Media markup (image, video, iframe…).
+  - description: Optional description rendered underneath the media.
+  - copyright: Optional configuration passed to the copyright overlay pattern. Provide an array
+    matching the copyright_overlay fields (content, trigger_label, modal_title, etc.).
+  - tag: HTML tag for the outer wrapper. Defaults to <figure>.
+  - attributes: Attributes for the outer wrapper.
+#}
+
+{%- set tag = tag|default('figure') %}
+{% set attributes = attributes|default(create_attribute()) %}
```
donquixote: I thought the whitespace for a single line break would disappear automatically.
The twig sandbox says otherwise.
But maybe it depends on the type of tag? Or it is different in Drupal?

Maybe we need this, but I don't know..

```twig
{%- set tag = tag|default('figure') %}
{%- set attributes = attributes|default(create_attribute()) %}
```

templates/patterns/media_container/pattern-media-container.html.twig
donquixote, line 1: Some thoughts about custom styles:

Introducing the additional `<div>` wrapper can interfere with some custom styles in projects.
But I don't see a way around it, so maybe that's just how it is.
Also, the problem will only start to occur when people start adding copyright.

Furthermore, when some items have a copyright and others do not, custom CSS needs to support both cases.

Not sure what to do about it, we might just have to accept it.

donquixote, line 1: One observation I made:
If the image or video does not fill up the full available width of the parent element, then the copyright overlay might not appear on top of the image but next to it.

Steps to reproduce:
- (I tried this in showcase where we have all the stuff available)
- Enable the "Standalone media URL" on /admin/config/media/media-settings.
- Add a copyright to an avportal video, visit that media page (desktop viewport).

Result: Video on the left, copyright on the right next to the video.

I don't know if this is something we can fix..
would be nice if CSS allowed to display relative to a sibling element. But it does not.

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -33,13 +45,13 @@
       {% endif %}
     {% endif %}
 
-    {%- if title is not empty -%}
+    {%- if title is not empty %}
       {% include '@oe-bcl/heading' with {
```
donquixote: I just did a little experiment in Drupal.
(just added the below snippet in a pattern template and visited that preview after cache clear)
The following line break will disappear:

```twig
<!-- begin test -->{% if true %}
{% endif %}<!-- end test -->
```
However, as soon as we introduce indentation, the line break still goes away, but the whitespace from the indentation remains:
```twig
<!-- begin test -->{% if true %}
  {% endif %}<!-- end test -->
```
So to standardize on having the "-" on the start of a twig tag:
```twig
    {%- if title is not empty %}
      {%- include '@oe-bcl/heading' with {
```

donquixote: See also https://github.com/twigphp/Twig/issues/4723

tests/src/Kernel/fixtures/PatternTestDataMassager.php
tibi2303: agree, we can simplify with your suggestion by a lot, please make a ticket.

donquixote: ok let's leave it like this for now

README.md
```diff
@@ -96,6 +96,27 @@ We started to ship in version 1.x patterns that already embrace this new approac
...
+
+1) Apply the recipes (they are not auto-applied):
+
+```bash
+# Run in your Drupal root where the theme is installed.
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_image_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_remote_video_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_av_portal_photo_copyright"
+./vendor/bin/drush recipe "/var/www/html/build/themes/contrib/oe_whitelabel/recipes/media_av_portal_video_copyright"
+```
```
donquixote: Shouldn't that documentation live in oe_whitelabel too? At least part of it?
We can mention it here for sure, but then?

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -50,9 +62,9 @@
           {% else %}
             {% set _subtitle_tag = title is not empty ? 'h3' : 'h2' %}
           {% endif %}
-          <{{ _subtitle_tag }} class="text-secondary">{{ subtitle }}</{{ _subtitle_tag }}>
+          <{{ _subtitle_tag }} class="text-secondary">{{- subtitle -}}</{{ _subtitle_tag }}>
```
donquixote: No space removal needed if there is no space in the template.

templates/patterns/media_container/pattern-media-container.html.twig
```diff
@@ -0,0 +1,44 @@
...
+          {{- pattern('copyright_overlay', _copyright) -}}
+        {% endif %}
+      </div>
+    {% endif %}
+    {% if has_description %}
+      <figcaption class="media-container__description">
+        {{- description -}}
+      </figcaption>
+    {% endif %}
+  </{{ tag }}>
```
donquixote: This outer layer with `<figure>` + `<figcaption>` seems not needed for what we do currently.
The feature request does not include a description.
Also, this could cause nested `<figure>` tag if the media itself already has that tag.

Just `<div>` with media + copyright seems enough.
If we do this change, we can name it something like "copyright media wrapper" rather than "media container".

We can still add something later to add a description.
But most of the components where media is displayed already have something where we would show a description. E.g. featured media pattern.

recipes/.gitignore
```diff
@@ -0,0 +1 @@
+/README.txt
```
donquixote: I think this file is auto-generated. But it is odd to have no newline at end of file.
If this is how it is generated then fine.

templates/patterns/gallery/pattern-gallery.html.twig
```diff
@@ -8,7 +8,8 @@
 
 {% set toggle_collapsed = 'View the full gallery (%d)'|t %}
 {% set toggle_expanded = 'View less'|t %}
-{% set items = bcl_gallery_items(items|default([])) %}
+{% set items = items|default([]) %}
+{% set items = bcl_gallery_items(items) %}
```
donquixote: why?

templates/patterns/modal/modal.ui_patterns.yml
```diff
@@ -2,6 +2,11 @@ modal:
   label: 'Modal'
   description: 'Use Bootstrap’s JavaScript modal plugin to add dialogs to your site for lightboxes, user notifications, or completely custom content. For more information vitis https://v5.getbootstrap.com/docs/5.0/components/modal/'
   settings:
+    vertically_centered:
+      type: boolean
+      label: 'Vertically centered'
+      description: 'Centers the modal in the viewport. Defaults to false.'
+      preview: false
```
donquixote: This is no longer in scope of the PR.

templates/patterns/modal/pattern-modal.html.twig
```diff
@@ -26,6 +26,7 @@
 {% include '@oe-bcl/modal' with {
   'size': size,
   'id': _id,
+  'vertically_centered': vertically_centered ?? false,
```
donquixote: No longer in scope because we don't use modal.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
```diff
@@ -137,6 +137,30 @@ private static function massageGalleryPattern(array $data): array {
...
+
```
donquixote: Does that pattern still exist?

README.md
donquixote: This comment still applies.
In oebt we would only document how to pass a copyright parameter to the templates.
(perhaps it needs no documentation at all in `README.md` because each pattern has its own documentation)
In whitelabel we can talk about recipes.

donquixote (changes requested):

templates/patterns/modal/modal.ui_patterns.yml
tibi2303: removed

templates/patterns/modal/pattern-modal.html.twig
tibi2303: removed.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
tibi2303: removed

recipes/.gitignore
tibi2303: yes, it's autogenerated on composer install like this.

templates/patterns/gallery/pattern-gallery.html.twig
tibi2303: rollbacked

recipes/README.txt
donquixote, line 1: If the `recipes/README.txt` is ignored via `recipes/.gitignore`, then why is it part of the PR?
(Typically this happens if you stage a file before it is ignored, see https://stackoverflow.com/questions/1274057/how-do-i-make-git-forget-about-a-file-that-was-tracked-but-is-now-in-gitignore

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -52,24 +53,24 @@
...
         {% endif %}
-        {{ text }}
+        {{- text -}}
         {% if link %}
           {{ pattern('link', {
             label: link.label,
             path: link.path,
             variant: 'link',
-            icon: link.icon,
+            icon: link.icon
```
donquixote: Let's keep that comma.
By having a trailing comma after last argument, when we add another argument it will be only one line of diff.
(this is why json is so annoying, you cannot do trailing comma)

templates/patterns/featured_media/pattern-featured-media.html.twig
```diff
@@ -52,24 +53,24 @@
           {% endif %}
           <{{ _subtitle_tag }} class="text-secondary">{{ subtitle }}</{{ _subtitle_tag }}>
         {% endif %}
-        {{ text }}
+        {{- text -}}
         {% if link %}
           {{ pattern('link', {
```
donquixote: Imo we don't need to touch this line in this PR.
If we do (in the future), we should follow a consistent and repeatable pattern like this:

```suggestion
        {{- text -}}
        {%- if link %}
          {{- pattern('link', {
```

This would be a pattern we can apply everywhere:
- If we need/want space removal between twig tags, we put the `-` at the beginning of the (next) twig tag.
- We put `-` at the end of a twig tag only if it is followed by whitespace and then html, not if it is followed by another twig tag.
- We could omit the `-` if all the html are block tags like `<div>`, that's up to us.

donquixote: (alternatively we could opt for `-%}` by default (right trim), it's a matter of convention.
I like left trim (`{%-`) because it is easily seen on the beginning of lines.

README.md
```diff
@@ -96,6 +96,12 @@ We started to ship in version 1.x patterns that already embrace this new approac
 We will do the best to avoid any breaking changes to these patterns during the 1.x lifecycle. But it might be that some changes will be still
 needed. We will communicate such changes together with the release.
 
+### Media copyright text
+
+Patterns that render media accept a `copyright` parameter to display copyright
+text under the media. See the pattern documentation for usage details. The
+field and recipe setup for media entities is documented in `oe_whitelabel`.
+
```
donquixote, line 104: Yep, this is short and sufficient!

recipes/README.txt
tibi2303: removed the recipes folder all together.

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: ok, i revert to {{ text }}

README.md
tibi2303: good, i will resolve

templates/patterns/featured_media/pattern-featured-media.html.twig
tibi2303: added back

donquixote (dismissed):

AaronGilMartinez (dismissed):

AaronGilMartinez (approved):

PR #528, opened 2026-01-06
OEL-4322: Migrate Alert and Description list components.

piotrsmykaj:


components/description_list/description_list.twig
```diff
@@ -0,0 +1,40 @@
...
+      {% set _term = _term|merge({
+        'icon': {'size': 'xs', 'path': _icon_path}|merge(_icon)
+      }) %}
+    {% endif %}
+
+    {% set _terms = _terms|merge([_term]) %}
+  {% endfor %}
+
+  {% set _items = _items|merge([item|merge({'term': _terms})]) %}
+{% endfor %}
```
donquixote, line 29: Could this be simplified by using `|map(..)`?
(on two levels)
The AI gives me this:
```twig
{% set _items = items|map(item =>
  item|merge({
    terms: (item.term is iterable ? item.term : [item.term])|map(term => {
      _term = term is iterable ? term : {'label': term}
    } => _term.icon is defined and _term.icon
      ? _term|merge({
          icon: {
            size: 'xs',
            path: _icon_path
          }|merge(_term.icon is iterable ? _term.icon : { name: _term.icon })
        })
      : _term
    )
  })
) %}
```
I am not saying this will work as-is, just as a starting point to explore.

donquixote: We could also create twig functions or filters if that helps. Either generic helpers, or more specific functions aimed at specific templates.
But not sure it would be worthwhile in this case.

includes/pattern.inc
```diff
@@ -59,39 +59,6 @@ function oe_bootstrap_theme_preprocess_pattern_offcanvas(array &$variables): voi
...
-        }
-        $term['icon'] += [
-          'size' => 'xs',
-          'path' => $variables['bcl_icon_path'],
-        ];
-      }
-    }
-  }
-}
-
```
donquixote, line 94: I see we are removing the preprocess, but the pattern itself is still there.

templates/overrides/menu/menu--main.html.twig
```diff
@@ -4,10 +4,10 @@
...
-      'attributes': attributes,
-    })
-}}
+{#{{#}
+{#    pattern('navigation', {#}
+{#      'variant': 'navbar',#}
+{#      'items': items,#}
+{#      'attributes': attributes,#}
+{#    })#}
+{#}}#}
```
donquixote, line 13: Is this necessary?
As a result, no menu will be shown?

We should find a different solution where we don't have to edit individual templates like this.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
```diff
@@ -71,10 +75,12 @@ private static function massageFilePattern(array $data): array {
...
    */
   private static function massageCardPattern(array $data): array {
-    if (isset($data['#fields']['image'])) {
-      $data['#fields']['image'] = ImageValueObject::fromArray($data['#fields']['image']);
+    if (!isset($data['#fields']['items'])) {
+      return $data;
     }
 
+    $data['#fields']['image'] = ImageValueObject::fromArray($data['#fields']['image']);
+
```
donquixote, line 83: Usually I like this early return pattern.
In this case I am not fully on board, because:
The method name `massageCardPatterns()` suggests that any alteration of the data for this pattern would happen in this method. For now it is only about items, but what if we want to also alter something else? Going by the method name, it would go into the same method.

```php
  private static function massageCardPattern(array $data): array {
    if (isset($data['#fields']['image'])) {
      $data['#fields']['image'] = ImageValueObject::fromArray($data['#fields']['image']);
    }
    if (isset($data['label'])) {
      $data['label'] = translate($data['label']);
    }
    return $data;
  }
```

It's not a hard blocker, I am just sharing my thoughts.

tests/src/Kernel/fixtures/PatternTestDataMassager.php
```diff
@@ -88,6 +94,10 @@ private static function massageCardPattern(array $data): array {
    *   The massaged data structure.
    */
   private static function massageCardLayoutPattern(array $data): array {
+    if (!isset($data['#fields']['items'])) {
+      return $data;
+    }
+
```
donquixote, line 100: Here the same arguments apply to the "early return", but I kind of like it because we don't have to touch the rest of the method for indentation :)

PR #535, opened 2026-01-16
OEL-4248: VRT Backstop

AaronGilMartinez:
Jira issue(s):


README.md
```diff
@@ -207,7 +207,22 @@ docker-compose exec web npm install
 docker-compose exec web npm run build
 # or, for continuous updates:
 docker-compose exec web npm run watch
-```
+
```
donquixote, line 215: You need to close the code snippet!
````suggestion
```

````

File: README.md [hunk 2]
```diff
@@ -209,6 +209,22 @@ docker-compose exec web npm run build
 docker-compose exec web npm run watch
 ```
 
+## Visual regression testing (BackstopJS)
+
+BackstopJS lives under `tests/vrt/`. The scenario builder and Playwright helper scripts in this directory generate `tests/vrt/scenarios/patterns.json` and `tests/vrt/backstop.generated.json`, which Backstop consumes for visual checks.
+
+### Prerequisites
+
+- A running Drupal site reachable at `http://localhost:8080/build`. Override the target with the `VRT_BASE_URL` environment variable if needed.
```
donquixote, line 222: I think we should simply point to the regular install steps.
The tests should pass on a fresh install following the steps above (docker-compose up, composer install, drupal:site-install).
There is no need to customize in the default case.
If you want to document these env vars, you could make a separate section about customization.

.github/workflows/backstop.yml
donquixote, line 1: We might want to check with QA if we can run this in gitlab.
But I am more than happy to first have it here in github, where we can control it all with code.

donquixote: Let's only start this conversation once we are happy with what we have.

README.md
```diff
@@ -209,6 +209,22 @@ docker-compose exec web npm run build
...
+- Valid credentials. The default `admin` / `admin` values can be overridden with `DRUPAL_USER` and `DRUPAL_PASS`.
+
+### Available commands
+
+- `docker compose exec web npm run vrt:build` – scrapes the UI Patterns catalog and generates `tests/vrt/scenarios/patterns.json` plus `tests/vrt/backstop.generated.json`.
+- `docker compose exec web npm run vrt:reference` – captures fresh reference screenshots into `tests/vrt/bitmaps_reference`.
+- `docker compose exec web npm run vrt:test` – compares the current UI against the references, writing results to `tests/vrt/bitmaps_test` and `tests/vrt/html_report`.
+- `docker compose exec web npm run vrt:approve` – promotes the latest passing screenshots to the reference set.
```
donquixote, line 231: When reading this, I would like to immediately understand in which category each of these commands fall, and when to run each of them:
1. I need to run this command as install / preparation before doing anything else.
1. I want to verify that the current appearance matches the captured snapshots.
2. I want to update the snapshots to align with changes in twig or CSS.

So here is what I understand:
- I always need to run `vrt:build` before anything else, because the two json files are not tracked in git. I also need to run this any time after I created, deleted or renamed patterns.
- I can run `vrt:reference` to update the snapshots after I changed the patterns.
- I can run `vrt:test` to see if snapshots match current appearance. As a side effect this creates some tmp files.
- I can run `vrt:approve` after `vrt:test`. Together, these achieve the same effect as `vrt:reference`. It is just a shortcut to avoid repeating the expensive calculation.

README.md
```diff
@@ -209,6 +209,22 @@ docker-compose exec web npm run build
...
+BackstopJS lives under `tests/vrt/`. The scenario builder and Playwright helper scripts in this directory generate `tests/vrt/scenarios/patterns.json` and `tests/vrt/backstop.generated.json`, which Backstop consumes for visual checks.
+
+### Prerequisites
+
+- A running Drupal site reachable at `http://localhost:8080/build`. Override the target with the `VRT_BASE_URL` environment variable if needed.
+- Valid credentials. The default `admin` / `admin` values can be overridden with `DRUPAL_USER` and `DRUPAL_PASS`.
+
+### Available commands
+
+- `docker compose exec web npm run vrt:build` – scrapes the UI Patterns catalog and generates `tests/vrt/scenarios/patterns.json` plus `tests/vrt/backstop.generated.json`.
```
donquixote, line 228: When I run this, it tell me to run `npx playwright install` first.
I also see that the github action does this.
We could add this to the regular install commands, perhaps using `runner.yml.dist`.

.github/workflows/backstop.yml
```diff
@@ -0,0 +1,61 @@
...
+
+      - name: Setup Node
+        uses: actions/setup-node@v6
+        with:
+          node-version: 24
+
+      - name: Install dependencies
+        run: |
+          npm install
+          npx playwright install --with-deps
```
donquixote, line 22: I notice that in `README.md` we run this npm stuff in web container, but here we run it in the root machine.

Nothing wrong, just an observation :)
(click "resolve" if this is as intended)

donquixote:
Rebased on 1.x.
Interestingly, github action only runs if there are no merge conflicts.
Gitlab tests run either way.

.github/workflows/backstop.yml
donquixote, line 22: We might want to split this step into two steps, simply to make it easier to see the output from each step in isolation.

README.md
```diff
@@ -209,6 +209,22 @@ docker-compose exec web npm run build
...
+- `docker compose exec web npm run vrt:reference` – captures fresh reference screenshots into `tests/vrt/bitmaps_reference`.
```
donquixote, line 229: I get a huge diff when I run this. But the images look the same.
Also, when I run it repeatedly, the files remain the same.
So maybe we just need to update once to make this PR green.

README.md
```diff
@@ -209,6 +209,22 @@ docker-compose exec web npm run build
...
+- `docker compose exec web npm run vrt:test` – compares the current UI against the references, writing results to `tests/vrt/bitmaps_test` and `tests/vrt/html_report`.
```
donquixote, line 230: I tried this.
How would I look at the output from this test, if there are failures? Which file do I look at?

README.md
donquixote, line 231: Can we run these commands targeting a specific snapshot or pattern?

.gitattributes
```diff
@@ -0,0 +1,60 @@
...
+*.eot     -text diff
+*.exe     -text diff
+*.gif     -text diff
+*.gz      -text diff
+*.ico     -text diff
+*.jpeg    -text diff
+*.jpg     -text diff
+*.otf     -text diff
+*.phar    -text diff
+*.png filter=lfs diff=lfs merge=lfs -text
```
donquixote, line 56: I played a bit with LFS myself.
This is great! Let's go with it.

PR #538, opened 2026-02-20
OEL-4032: Fix button pattern.

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/button/pattern-button.html.twig
+++ b/templates/patterns/button/pattern-button.html.twig
@@ -4,7 +4,6 @@
  * Button pattern.
  */
 #}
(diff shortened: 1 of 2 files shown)


templates/patterns/button/pattern-button.html.twig
```diff
@@ -4,11 +4,10 @@
  * Button pattern.
  */
 #}
+{% set _icon = null %}
```
AaronGilMartinez: No need for this, conditionals dont create their own scopes, so _icon will be available once its set inside the if.

AaronGilMartinez (approved):

PR #541, opened 2026-03-03
OEL-4511: Update to Drupal 11.3.x.

tibi2303:


theme-settings.php
```diff
@@ -9,11 +9,14 @@
 
 use Drupal\Core\Form\FormStateInterface;
 use Drupal\oe_bootstrap_theme\BackwardCompatibility;
+use Drupal\oe_bootstrap_theme\ThemeSettings;
 
 /**
  * Implements hook_form_system_theme_settings_alter().
  */
 function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateInterface $form_state) {
+  $theme_name = 'oe_bootstrap_theme';
```
donquixote: That local variable seems unnecessary.

src/ThemeSettings.php
```diff
@@ -0,0 +1,40 @@
...
+   *   The theme name. If omitted, the active theme is used first and the
+   *   default theme as fallback.
+   *
+   * @return mixed
+   *   The setting value.
+   */
+  public static function get(string $setting, ?string $theme = NULL): mixed {
+    if (version_compare(\Drupal::VERSION, self::THEME_SETTINGS_PROVIDER_VERSION, '>=')) {
+      $service_id = 'Drupal\Core\Extension\ThemeSettingsProvider';
+      $settings_provider = \Drupal::service($service_id);
```
donquixote:
```suggestion
      $settings_provider = \Drupal::service(Drupal\Core\Extension\ThemeSettingsProvider::class);
```
- no need for local var
- use ::class so that static analysis sees the class name.

src/ThemeSettings.php
```diff
@@ -0,0 +1,40 @@
...
+      return \call_user_func([$settings_provider, 'getSetting'], $setting, $theme);
```
donquixote: why call_user_func?
```suggestion
      return $settings_provider->getSetting($setting, $theme);
```
If you are concerned about static analysis, you can put a `/** @var */` doc on the service.

src/ThemeSettings.php
donquixote: and then use class import
```php
use Drupal\Core\Extension\ThemeSettingsProvider;
```

```suggestion
      $settings_provider = \Drupal::service(ThemeSettingsProvider::class);
```

src/ThemeSettings.php
donquixote: if you don't need the local var for the service:
```suggestion
      return \Drupal::service(ThemeSettingsProvider::class)->getSetting($setting, $theme);
```
The local var for the service is useful only if you want to put a `/** @var */` doc.

src/ThemeSettings.php
```diff
@@ -0,0 +1,40 @@
...
+    }
+
+    $legacy_getter = 'theme_get_setting';
+    return \call_user_func($legacy_getter, $setting, $theme);
```
donquixote:
```suggestion
    // phpcs:ignore <deprecation code>
    return theme_get_setting($setting, $theme);
```

not sure how that phpcs line needs to look like or if you need it at all.

theme-settings.php
```diff
@@ -25,7 +28,7 @@ function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateIn
     '#type' => 'checkbox',
     '#title' => t('Style all tables using Bootstrap.'),
     '#description' => t('Applies Bootstrap classes to all tables rendered using this theme. Note that some table instances (such as calendars) might not render correctly when Bootstrap is applied to them.'),
-    '#default_value' => theme_get_setting('bootstrap_tables.enable'),
+    '#default_value' => ThemeSettings::get('bootstrap_tables.enable', $theme_name),
```
donquixote:
```suggestion
    '#default_value' => ThemeSettings::get('bootstrap_tables.enable', 'oe_bootstrap_theme'),
```

But do you even need to pass the theme name at all?
Shouldn't it work without that?

```suggestion
    '#default_value' => ThemeSettings::get('bootstrap_tables.enable'),
```

This also helps to support overriding the setting in a subtheme, which I think was the goal of  theme_get_setting().

src/ThemeSettings.php
donquixote, line 1: Alternatively we could name this class `DrupalCompatibility` and the method `themeGetSetting()`, this was suggested by AI :)

The point would be to add more like this in the future.

it's a bit confusing with the existing class BackwardCompatibility, which is meant not to support old Drupal versions but to solve another kind of problem (not sure how that should work)

donquixote (changes requested):
some feedback.

I am still undecided if it would be acceptable to simply accept or suppress the deprecated calls. I am trying to find the discussion in drupal.org where this was mentioned.

EDIT: No, we don't suppress :)
(resolved)

src/ThemeSettings.php
```diff
@@ -0,0 +1,40 @@
...
+  }
```
donquixote: Actually, something better!
https://www.drupal.org/node/3379306
Can you try this?

```php
use Drupal\Component\Utility\DeprecationHelper;
[...]
    return DeprecationHelper::backwardsCompatibleCall(
      \Drupal::VERSION,
      '11.3',
      \Drupal::service(ThemeSettingsProvider::class)->getSetting(...),
      theme_get_setting(...),
    );
```

oe_bootstrap_theme.theme
```diff
@@ -121,10 +122,23 @@ function oe_bootstrap_theme_preprocess_views_view_table__bootstrap__responsive(&
...
+ *
+ * @param string $setting
+ *   The setting name.
+ *
+ * @return mixed
+ *   The setting value.
+ */
+function _oe_bootstrap_theme_get_setting(string $setting): mixed {
+  return ThemeSettings::get($setting, 'oe_bootstrap_theme');
+}
```
donquixote: Why this function, instead of calling the static method directly?

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+        $settings_provider = \Drupal::service('Drupal\Core\Extension\ThemeSettingsProvider');
+        if (is_object($settings_provider) && method_exists($settings_provider, 'getSetting')) {
+          return $settings_provider->getSetting($setting, $theme);
+        }
+
+        throw new \RuntimeException('ThemeSettingsProvider::getSetting() is not available.');
+      },
+      static function () use ($setting, $theme): mixed {
+        return \theme_get_setting($setting, $theme);
+      }
```
donquixote: we can simplify with `fn ()`. Also use trailing comma on last arg.

```suggestion
      fn () =>\theme_get_setting($setting, $theme),
```
(and same for the other one)

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+      \Drupal::VERSION,
+      self::THEME_SETTINGS_PROVIDER_VERSION,
+      static function () use ($setting, $theme): mixed {
+        $settings_provider = \Drupal::service('Drupal\Core\Extension\ThemeSettingsProvider');
+        if (is_object($settings_provider) && method_exists($settings_provider, 'getSetting')) {
+          return $settings_provider->getSetting($setting, $theme);
+        }
+
+        throw new \RuntimeException('ThemeSettingsProvider::getSetting() is not available.');
+      },
```
donquixote: No need for all the defensive code.
If we have the correct Drupal version, it will just work. The service won't surprisingly be something we did not expect.
If it is, we will get some kind of error and we will know.
Also always use ::class, not string literal with quotes for a class name. This way an IDE can do "Find usages" etc.

```suggestion
      fn () => \Drupal::service(ThemeSettingsProvider::class)->getSetting($setting, $theme),
```

If there is any problem with phpcs or phpstan when running a different core version, we suppress.

donquixote: As a general note: Defensive code is more important when we deal with user data, or with anything that could have different values or states for reasons we do not control.

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,40 @@
...
+   *
+   * @return mixed
+   *   The setting value.
+   */
+  public static function themeGetSetting(string $setting, ?string $theme = NULL): mixed {
+    return DeprecationHelper::backwardsCompatibleCall(
+      \Drupal::VERSION,
+      self::THEME_SETTINGS_PROVIDER_VERSION,
+      fn () => \Drupal::service(ThemeSettingsProvider::class)->getSetting($setting, $theme),
+      fn () =>\theme_get_setting($setting, $theme),
```
donquixote: We need a whitespace here.
And the `\` backslash is debatable, we typically don't add it.

```suggestion
      fn () => theme_get_setting($setting, $theme),
```

oe_bootstrap_theme.theme
```diff
@@ -41,13 +42,13 @@ function oe_bootstrap_theme_theme_suggestions_table_alter(array &$suggestions, a
...
 function oe_bootstrap_theme_theme_suggestions_views_view_table_alter(array &$suggestions, array $variables) {
-  if (!theme_get_setting('bootstrap_tables.enable')) {
+  if (!DrupalCompatibility::themeGetSetting('bootstrap_tables.enable', 'oe_bootstrap_theme')) {
     return;
   }
 
   $extra_suggestions[] = 'views_view_table__bootstrap';
 
-  $responsive = theme_get_setting('bootstrap_tables.responsive');
+  $responsive = DrupalCompatibility::themeGetSetting('bootstrap_tables.responsive', 'oe_bootstrap_theme');
```
donquixote: Again, why pass the theme name?

I assume we want the following to work:
- You enable oe_whitelabel, which is a subtheme of oe_bootstrap_theme.
- In oe_whitelabel, you have the same config options available (and more).
- Setting values are stored separately for oe_whitelabel and oe_bootstrap_theme.
- When this function is called while in oe_whitelabel, we want the setting value from oe_whitelabel to be used.


```suggestion
  $responsive = DrupalCompatibility::themeGetSetting('bootstrap_tables.responsive');
```

donquixote: I don't fully remember if this is how these theme settings work, please try it out.
But there is a reason why until now we did not pass along the theme name.

src/BackwardCompatibility.php
```diff
@@ -27,7 +32,7 @@ final class BackwardCompatibility {
    * @SuppressWarnings(PHPMD.BooleanGetMethodName)
    */
   public static function getSetting(string $name): bool {
-    return theme_get_setting(self::PREFIX . $name) ?? TRUE;
+    return (bool) (DrupalCompatibility::themeGetSetting(self::PREFIX . $name, self::THEME) ?? TRUE);
```
donquixote: We are still passing the theme name?

donquixote:
```suggestion
    return (bool) (DrupalCompatibility::themeGetSetting(self::PREFIX . $name) ?? TRUE);
```

src/BackwardCompatibility.php
```diff
@@ -14,6 +14,11 @@ final class BackwardCompatibility {
    */
   public const PREFIX = 'backward_compatibility.';
 
+  /**
+   * The theme where backward compatibility settings are stored.
+   */
+  private const THEME = 'oe_bootstrap_theme';
+
```
donquixote: should not be needed.

theme-settings.php
```diff
@@ -25,7 +26,7 @@ function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateIn
     '#type' => 'checkbox',
     '#title' => t('Style all tables using Bootstrap.'),
     '#description' => t('Applies Bootstrap classes to all tables rendered using this theme. Note that some table instances (such as calendars) might not render correctly when Bootstrap is applied to them.'),
-    '#default_value' => theme_get_setting('bootstrap_tables.enable'),
+    '#default_value' => DrupalCompatibility::themeGetSetting('bootstrap_tables.enable', 'oe_bootstrap_theme'),
```
donquixote:
```suggestion
    '#default_value' => DrupalCompatibility::themeGetSetting('bootstrap_tables.enable'),
```

theme-settings.php
```diff
@@ -41,7 +42,7 @@ function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateIn
       'xxl' => t('Extra extra large'),
     ],
     '#empty_option' => t('Never'),
-    '#default_value' => theme_get_setting('bootstrap_tables.responsive') ?? '',
+    '#default_value' => DrupalCompatibility::themeGetSetting('bootstrap_tables.responsive', 'oe_bootstrap_theme') ?? '',
```
donquixote:
```suggestion
    '#default_value' => DrupalCompatibility::themeGetSetting('bootstrap_tables.responsive') ?? '',
```

tibi2303: done

theme-settings.php
tibi2303: done

src/BackwardCompatibility.php
tibi2303: removed

theme-settings.php
```diff
@@ -25,7 +26,7 @@ function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateIn
     '#type' => 'checkbox',
     '#title' => t('Style all tables using Bootstrap.'),
     '#description' => t('Applies Bootstrap classes to all tables rendered using this theme. Note that some table instances (such as calendars) might not render correctly when Bootstrap is applied to them.'),
-    '#default_value' => theme_get_setting('bootstrap_tables.enable'),
+    '#default_value' => DrupalCompatibility::themeGetSetting('bootstrap_tables.enable'),
```
donquixote, line 29: It seems we need to remove this line for correct behavior. Drupal ConfigFormBase will auto fill the setting. To be confirmed if this works in Drupal 10, I think it should.

```suggestion
```

donquixote: Compare how core themes do it, e.g. olivero/theme-settings.php

theme-settings.php
```diff
@@ -25,7 +26,6 @@ function oe_bootstrap_theme_form_system_theme_settings_alter(&$form, FormStateIn
     '#type' => 'checkbox',
     '#title' => t('Style all tables using Bootstrap.'),
     '#description' => t('Applies Bootstrap classes to all tables rendered using this theme. Note that some table instances (such as calendars) might not render correctly when Bootstrap is applied to them.'),
-    '#default_value' => theme_get_setting('bootstrap_tables.enable'),
```
donquixote, line 28: I am sorry for my dead-end and wrong advice earlier.
I did some testing and find that `DrupalCompatibility::themeGetSetting($config_name)` (without another argument) should just work here.

I tried this with both 11.3 and 11.2.

The reason this works is that `ThemeSettingsForm` temporarily sets the active theme to whichever theme we are editing, before calling the alter hook. This way, any call to `theme_get_setting()` or its Drupal 11.3 equivalent will get the value from the theme being edited.

I also found that `olivero_form_system_theme_settings_alter()` is doing it wrong.
When you enable a sub-theme of olivero, and edit the settings, they are saved to olivero rather than the sub-theme.

donquixote: Another interesting fact: ThemeSettingsForm extends ConfigFormBase, which has its own mechanism to read form values and write them to config on submit. To do that, it looks at the `#config_target` keys of form elements.

ThemeSettingsForm calls the parent submit method, but that is completely pointless, because generally none of the elements have #config_target.
Instead, it then does its own mechanism to write values to config, grabbing all form values instead of just those where #config_target was set, and writing to the correct theme.

Olivero sets #config_target to e.g. olivero.settings:site_branding_bg_color, which is wrong.
As a consequence, all values are loaded from olivero, then saved to olivero on submit, then saved to the sub-theme (I think).
It is ok because olivero is not designed as a base theme.

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,42 @@
...
+
+/**
+ * Handles Drupal core compatibility behaviors.
+ */
+final class DrupalCompatibility {
+
+  /**
+   * Service id for the Drupal 11.3+ theme settings provider.
+   */
+  private const THEME_SETTINGS_PROVIDER_SERVICE = 'Drupal\\Core\\Extension\\ThemeSettingsProvider';
```
donquixote: There is no need for this constant.
Also we should import the class and use `ThemeSettingsProvider::class`.
Actually the earlier version was fine, where we had `\Drupal::service(ThemeSettingsProvider::class)`.
Or was there a problem?

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,42 @@
...
+
+  /**
+   * Core version where ThemeSettingsProvider::getSetting() should be used.
+   */
+  private const THEME_SETTINGS_PROVIDER_VERSION = '11.3.0';
```
donquixote: I don't see a value of having this in a constant.
It makes much more sense to have all the info related to `ThemeSettingsProvider` / `theme_get_settings()` centralized in the method.

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,32 @@
...
+   *   The setting name.
+   *
+   * @return mixed
+   *   The setting value.
+   */
+  public static function themeGetSetting(string $setting): mixed {
+    return DeprecationHelper::backwardsCompatibleCall(
+      \Drupal::VERSION,
+      '11.3.0',
+      fn () => \Drupal::service('Drupal\\Core\\Extension\\ThemeSettingsProvider')->getSetting($setting),
```
donquixote: Import and use `::class`!
```php
use Drupal\Core\Extension\ThemeSettingsProvider;
```

```suggestion
      fn () => \Drupal::service(ThemeSettingsProvider::class)->getSetting($setting),
```

tibi2303: removed.

theme-settings.php
donquixote: Olivero was a dead end, sorry.
But now it is good.

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+   *   The setting name.
+   *
+   * @return mixed
+   *   The setting value.
+   */
+  public static function themeGetSetting(string $setting): mixed {
+    return DeprecationHelper::backwardsCompatibleCall(
+      \Drupal::VERSION,
+      '11.3.0',
+      fn () => \Drupal::service('Drupal\Core\Extension\ThemeSettingsProvider')->getSetting($setting),
```
donquixote, line 27: Please import the class and use `::class`!

tests/src/Functional/ThemeSettingsTest.php
```diff
@@ -216,7 +212,7 @@ protected function renderTestTable(string $theme = 'table'): string {
 
     // Theme settings config is cached statically.
     drupal_static_reset();
-    \Drupal::configFactory()->clearStaticCache();
+    DrupalCompatibility::resetThemeSettingsCache();
```
donquixote, line 219: Do we still need the `drupal_static_reset();` above?
(I don't have the answer, you need to try)

src/DrupalCompatibility.php
tibi2303: image
It does not pass the review, check please the previous commit, that one that does not pass...

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+   * Resets the theme settings cache across supported Drupal core versions.
+   *
+   * In Drupal >= 11.3, ThemeSettingsProvider uses an in-memory cache that must
+   * be cleared separately from the config factory static cache.
+   */
+  public static function resetThemeSettingsCache(): void {
+    DeprecationHelper::backwardsCompatibleCall(
+      \Drupal::VERSION,
+      '11.3.0',
+      fn () => \Drupal::service('cache.memory')->deleteAll(),
```
donquixote: Here is the commit where they deprecated `theme_get_setting()`: https://git.drupalcode.org/project/drupal/-/commit/be276ecb985c282019f5c0747b745510555c5c5f
Change record: https://www.drupal.org/node/3035289
Issue: https://www.drupal.org/project/drupal/issues/3035288

The replacement in the PR for `drupal_static_reset('theme_get_setting')` was not a full deletion of all memory cache, but invalidation of specific cache tags.
```php
\Drupal::service('cache.memory')->invalidateTags(['config:stark.settings']);
```
In the change record, we see a different solution:
```php
// Clear the 'bartik' theme settings static cache.
\Drupal::service('cache.memory')->delete('theme_settings:bartik');

// Clear the theme settings static cache for all themes.
\Drupal::service('cache_tags.invalidator')->invalidateTags(['config:system.theme.global']);
```
(I commented under the change record, asking for clarification.)


Now the tricky bit:
The `drupal_static_reset()` worked for all themes.
The solution that was applied in Drupal core and in the first suggestion targets a specific theme.
In our case that would be `oe_bootstrap_theme`. But we also want to be able to use this method in `oe_whitelabel`.

So I think we should go for the "for all theme", and see if that works.

, but the new line targets settings for a specific theme only (in that case 'stark').
In this PR, we are only using this in tests for oe_bootstrap_theme, so it would be ok to hard-code that theme name.
But we might also want to use the method in sub-themes like `oe_whitelabel`. I looked at oe_whitelabel code, I also see it only called in tests.

So we could do something like this:
```php
  /**
   * Resets the theme settings memory cache across supported Drupal versions.
   *
   * This is needed in functional tests, when a separate request made changes to
   * configuration, but the old version is still in the memory cache of the
   * process where that test is running.
   *
   * In Drupal < 11.3, this memory cache was using the drupal_static()
   * mechanism. In Drupal >= 11.3, the new memory cache service is used.
   *
   * See https://www.drupal.org/node/3035289.
   */
  public static function resetThemeSettingsCache(): void {
    DeprecationHelper::backwardsCompatibleCall(
      \Drupal::VERSION,
      '11.3.0',
      fn () => \Drupal::service('cache_tags.invalidator')->invalidateTags(['config:system.theme.global']),
      fn () => drupal_static_reset('theme_get_setting'),
    );
  }
```

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+      fn () => drupal_static_reset('theme_get_setting'),
+    );
+    \Drupal::configFactory()->clearStaticCache();
```
donquixote: For this line:
I understand the desire to centralize, given this is repeated in so many places.
But it is actually a separate thing, has nothing to do with the BC support.

I see two options:
1. Leave the call in place everywhere, only replace the drupal_static_reset().
2. In addition to this BC method, provide a separate method outside of this class to clear both the config factory static cache and the theme settings.

I would go for option 1 to avoid unnecessary changes which do not need to be part of this PR.

tibi2303: Applied option 1. The compatibility helper now only handles the theme-settings cache reset for Drupal 11.3+, and the config factory static cache clear remains explicit in the test call sites.

src/DrupalCompatibility.php
tibi2303: redone as you said

src/DrupalCompatibility.php
```diff
@@ -0,0 +1,48 @@
...
+  }
```
donquixote, line 46: Actually.. regarding my comments above:
If we rename this method to `clearAllMemoryCaches()`, and then really do that (full drupal_static_reset(), clear config factory static cache, clear all of cache.memory), then it would make sense to have it all in one method (but maybe in a separate class because it is not strictly about BC?).

Normally in a test or elsewhere we only want to clear or invalidate as little as possible, because we want to be sure that the regular test invalidation works.

Here the reason why we would just clear all memory caches is that we have a classical problem where the change happens in a different process, and the process that runs the test cannot know about it. So naturally all the memory caches will become outdated. This applies to everything from drupal_static(), the cache.memory, the config static cache, but also other stuff like entities cached in an entity storage property.

So.. imo let's go for the minimal change solution where we only replace drupal_static_reset() and nothing else :)

src/DrupalCompatibility.php
donquixote: Alright.
There are ways to have different PhpStan ignore rules per Drupal core version.
But this is way too much for now, so let's keep that simple solution with the string literal.

src/DrupalCompatibility.php
donquixote: Alright, I see this is what you did in the end - just replace drupal_static_reset(), keep the rest.
-> ok for me.

tibi2303: `DrupalCompatibility::resetThemeSettingsCache()` only replaces the old
`drupal_static_reset('theme_get_setting')` behavior for Drupal 11.3+.
The config factory static cache clear stays explicit at the test call sites,
so I did not broaden this to clear anything else.

donquixote (approved):

PR #544, opened 2026-03-24
OEL-4562: Fix rich text rendering in listing item block paragraphs.

tibi2303:
Jira issue(s):


modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -139,7 +140,7 @@ public function bclCardList(array $items): array {
       }
       if (isset($item['text'])) {
         $bcl_card['text'] = [
-          'content' => $item['text'],
+          'content' => $this->normalizeCardTextContent($item['text']),
```
donquixote, line 143: Why can't the calling template pass a markup object as content?
This should be the general contract for all components.

PR #547, opened 2026-03-29
OEL-4500: Migrate description_list to SDC component

piotrsmykaj:


tests/src/PatternAssertion/BasePatternAssert.php
donquixote, line 1: We _could_ rethink this pattern assert mechanism, but for now it is good to keep it aligned and not change too much, so +1 for this solution.

tests/src/Kernel/ComponentRenderingTest.php
donquixote, line 1: I was wondering if we should reuse the MarkupRenderingTest and have it test components too.
But I think it is good like this to have a new class, even if a lot of this is copied.
So +1.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -57,11 +73,17 @@ class TwigExtension extends AbstractExtension {
    *   The Drupal Twig environment.
    * @param \Drupal\Core\Render\RendererInterface $renderer
    *   The renderer.
+   * @param \Drupal\Core\Extension\ThemeExtensionList $theme_list
+   *   The theme extension list.
+   * @param \Drupal\Core\Theme\ThemeManagerInterface $theme_manager
+   *   The theme manager.
    */
-  public function __construct(LanguageManagerInterface $languageManager, TwigEnvironment $twigEnvironment, RendererInterface $renderer) {
+  public function __construct(LanguageManagerInterface $languageManager, TwigEnvironment $twigEnvironment, RendererInterface $renderer, ThemeExtensionList $theme_list, ThemeManagerInterface $theme_manager) {
```
donquixote: Can we do autowire and constructor property promotion?

donquixote: Also we can omit doc comment on constructor, if there is no unusual parameter.

components/description_list/description_list.twig
```diff
@@ -0,0 +1,40 @@
...
+      {% set _term = _term|merge({
+        'icon': {'size': 'xs', 'path': _icon_path}|merge(_icon)
+      }) %}
+    {% endif %}
+
+    {% set _terms = _terms|merge([_term]) %}
+  {% endfor %}
+
+  {% set _items = _items|merge([item|merge({'term': _terms})]) %}
+{% endfor %}
```
donquixote: I would like if we can keep this data processing logic out of the template.
Either using a twig function, or by having a more narrow schema for the input values.

components/description_list/description_list.component.yml
```diff
@@ -0,0 +1,76 @@
...
+      title: 'List items'
+      type: array
+      description: 'Each item is composed by one or more terms and one or more definitions. Each term entry can have an icon.'
+      items:
+        type: object
+        properties:
+          term:
+            title: 'Terms'
+            type: [string, array]
+            description: 'Can be a single string or an array of term objects.'
```
donquixote: Do we want to support this multimodal input?
Or should we force one format? I think I would prefer that.

components/description_list/description_list.component.yml
donquixote, line 1: Is the props schema validated somewhere?
E.g. if the test data does not match the schema, would we know from a failing test?

components/description_list/description_list.component.yml
piotrsmykaj: I do agree, better to keep types for icon and term singular and clean. Nevertheless, I would keep types: [string, array, object] for title and items label to allow a text input in any of these types.

components/description_list/description_list.twig
piotrsmykaj: ok good ideas. I will do both things.

components/description_list/description_list.component.yml
piotrsmykaj: Good question. I checked it, and it fails if the value type doesn't match the schema type. We validate it because the "zend.assertions" PHP setting is set to 1 (both in -dev and -ci).

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -27,49 +29,21 @@
...
-    $this->twigEnvironment = $twigEnvironment;
-    $this->renderer = $renderer;
-  }
+  public function __construct(
+    protected LanguageManagerInterface $languageManager,
+    protected TwigEnvironment $twigEnvironment,
+    protected RendererInterface $renderer,
+    protected ThemeExtensionList $themeList,
+    protected ThemeManagerInterface $themeManager,
+  ) {}
```
donquixote, line 38: I'd say most of it can be readonly - if the service is not going to be serialized.

```suggestion
  public function __construct(
    protected readonly LanguageManagerInterface $languageManager,
    protected readonly TwigEnvironment $twigEnvironment,
    protected readonly RendererInterface $renderer,
    protected readonly ThemeExtensionList $themeList,
    protected readonly ThemeManagerInterface $themeManager,
  ) {}
```

We _could_ also declare the class itself as readonly, but I don't like that:
As soon as we add any one property that is not readonly (e.g. for some kind of cache), we have to remove the `readonly` from the class and have to re-add it to the properties. So I only put it on properties.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -164,6 +139,33 @@ public function bclCardList(array $items): array {
...
+   * @return array
+   *   The processed items.
+   */
+  public function bclDescriptionListItems(array $items, ?string $icon_path = NULL): array {
+    if (!$icon_path) {
+      $icon_path = $this->getBclIconPath();
+    }
+
+    foreach ($items as &$item) {
+      foreach ($item['term'] as &$term) {
```
donquixote, line 165: Is `$item['term']` always present?
I wrote elsewhere about a more narrow schema, but I don't know how narrow we should make it.

components/description_list/description_list.component.yml
```diff
@@ -0,0 +1,76 @@
...
+                  properties:
+                    name:
+                      type: string
+                    size:
+                      type: string
+                    path:
+                      type: string
+          definition:
+            title: 'Definition'
+            type: [string, array]
```
donquixote: We are allowing string and array, but then we have `items:` which only works for array.

Maybe it would be cleaner with `oneOf:` (or `anyOf:`). But I don't know if these are supported in a *.component.yml. This would have to be tested.

I don't find it anywhere in core component yml files.
I do find `AtLeastOneOf:` in some core yaml files but not for components.
I do find `"anyOf":` and `"oneOf":` in some json schema files in core.

components/description_list/description_list.story.twig
```diff
@@ -0,0 +1,41 @@
...
+  items: [
+    {
+      term: [
+        {
+          label: 'Single label with icon',
+          icon: { name: 'geo-alt-fill' }
+        }
+      ],
+      definition: 'Description of first term text goes here.'
+    },
```
donquixote, line 12: We are now allowing `string|array` for definition, but require array for term. This feels inconsistent.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
donquixote: In the BCL template, 'term' seems to be optional.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -164,6 +139,33 @@ public function bclCardList(array $items): array {
...
+        if (!empty($term['icon'])) {
+          $term['icon'] += ['size' => 'xs', 'path' => $icon_path];
+        }
+      }
+    }
+
+    return $items;
+  }
```
donquixote, line 175: For now I think this is good, and I prefer seeing this logic in php over seeing it in twig.
Maybe at some point we can get ideas to have reusable twig functions that can be used across patterns.

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
piotrsmykaj: ah true

components/description_list/description_list.component.yml
piotrsmykaj: i think "items:" are ignored for "string" type, and on the BCL level, both types are handled for both properties:

```
{%- if _item.term is iterable %}
  {%- for _term in _item.term %}
    {%- include '@oe-bcl/bcl-description-list/description-list-title.html.twig' with _term only -%}
  {% endfor %}
{%- else %}
  {%- include '@oe-bcl/bcl-description-list/description-list-title.html.twig' with {
    label: _item.term
  } only -%}
{% endif %}

[...]

{%- if _item.definition is iterable %}
  {%- for _definition in _item.definition %}
    {%- include '@oe-bcl/bcl-description-list/description-list-details.html.twig' with _definition only -%}
  {% endfor %}
{%- else %}
  {%- include '@oe-bcl/bcl-description-list/description-list-details.html.twig' with {
    label: _item.definition
  } only -%}
{% endif %}
```

but I checked that anyOf works in SDC, so I will use it for both properties to make it consistent.

components/description_list/description_list.story.twig
piotrsmykaj: see the comment above

components/description_list/description_list.component.yml
```diff
@@ -0,0 +1,81 @@
...
+            description: 'A string or an array of term objects.'
+            anyOf:
+              - type: string
+              - type: array
+                items:
+                  type: object
+                  properties:
+                    label:
+                      title: Label
+                      type: [string, array, object]
```
donquixote, line 32: It is sad there is no dedicated meta-type for `MarkupInterface|string|render-array`

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -164,6 +139,36 @@ public function bclCardList(array $items): array {
...
+   * @return array
+   *   The processed items.
+   */
+  public function bclDescriptionListItems(array $items, ?string $icon_path = NULL): array {
+    if (!$icon_path) {
+      $icon_path = $this->getBclIconPath();
+    }
+
+    foreach ($items as &$item) {
+      if (empty($item['term']) || !is_array($item['term'])) {
```
donquixote: If `$item['term']` is not an array but not empty, what else could it be?
We either want to fail or support that case.

To "fail" not sure what is the correct exception type for twig functions. In `Drupal\Core\Template\TwigExtension` I see regular `\Exception` being thrown.

piotrsmykaj: > If $item['term'] is not an array but not empty, what else could it be?

It can be the string.

> In Drupal\Core\Template\TwigExtension I see regular \Exception being thrown.

[EA] \Exception is too general. Consider throwing one of SPL exceptions instead. 
I used the "\InvalidArgumentException"

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -164,6 +139,39 @@ public function bclCardList(array $items): array {
...
+   *
+   * @return array
+   *   The processed items.
+   */
+  public function bclDescriptionListItems(array $items, ?string $icon_path = NULL): array {
+    if (!$icon_path) {
+      $icon_path = $this->getBclIconPath();
+    }
+
+    foreach ($items as &$item) {
```
donquixote, line 158: Btw I am used to seeing critical feedback about by-reference variables in `foreach ()`.
There are definitely some risks if not used carefully. But I personally don't have a problem with it as much as some others do - so I accept it :)

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
```diff
@@ -164,6 +139,39 @@ public function bclCardList(array $items): array {
...
+      if (empty($item['term']) || is_string($item['term'])) {
+        continue;
+      }
```
donquixote, line 161: So what is the expected shape of the `$items` in the end?
With the `continue;` above, the result can then have some items with a string for 'term', and others with an array.

donquixote: actually 'term' being a string will be ok because bcl takes care of it - right?

piotrsmykaj: > actually 'term' being a string will be ok because bcl takes care of it - right?

yes

modules/oe_bootstrap_theme_helper/src/TwigExtension/TwigExtension.php
piotrsmykaj: ok good point, but currently, directly after `foreach()`, there is the return. I added unset() for $term and $item to mitigate the risk in the future.

donquixote:
test comment, please ignore :)

components/description_list/description_list.component.yml
donquixote: another test comment, please ignore :)

donquixote (approved):

PR #549, opened 2026-04-01
OEL-4500: Backwards compatible description list component.

AaronGilMartinez:
Jira issue(s):

### DIFF
--- a/components/description_list/description_list.twig
+++ b/components/description_list/description_list.twig
@@ -7,7 +7,7 @@
 {% include '@oe-bcl/description-list' with {
   title: title,
   title_link: title_link,
-  title_tag: title_tag|default('h2'),
+  title_tag: title_tag,
   title_attributes: title_attributes,
   items: items|bcl_description_list_items(icon_path),
   variant: orientation,
--- a/templates/patterns/description_list/pattern-description-list.html.twig
+++ b/templates/patterns/description_list/pattern-description-list.html.twig
@@ -4,10 +4,10 @@
  * Description list pattern.
  */
 #}
-{% include '@oe-bcl/description-list' with {
-  'title': title,
-  'title_tag': title_tag,
-  'items': items,
-  'variant': orientation,
-  'attributes': attributes,
-} only %}
+{% embed 'oe_bootstrap_theme:description_list' with {
+  title: title,
+  title_tag: title_tag|default('h2'),
+  items: items,
+  orientation: orientation,
+  attributes: attributes,
+} only %}{% endembed %}
--- a/templates/patterns/description_list/pattern-description-list.html.twig
+++ b/templates/patterns/description_list/pattern-description-list.html.twig
@@ -6,7 +6,7 @@
 #}
 {% embed 'oe_bootstrap_theme:description_list' with {
   title: title,
-  title_tag: title_tag|default('h2'),
+  title_tag: title_tag,
   items: items,
   orientation: orientation,
   attributes: attributes,
--- a/templates/patterns/description_list/pattern-description-list.html.twig
+++ b/templates/patterns/description_list/pattern-description-list.html.twig
@@ -4,10 +4,10 @@
  * Description list pattern.
  */
 #}
-{% embed 'oe_bootstrap_theme:description_list' with {
+{% include 'oe_bootstrap_theme:description_list' with {
   title: title,
   title_tag: title_tag,
   items: items,
   orientation: orientation,
   attributes: attributes,
-} only %}{% endembed %}
+} only %}


templates/patterns/description_list/pattern-description-list.html.twig
```diff
@@ -4,10 +4,10 @@
...
-{% include '@oe-bcl/description-list' with {
-  'title': title,
-  'title_tag': title_tag,
-  'items': items,
-  'variant': orientation,
-  'attributes': attributes,
-} only %}
+{% embed 'oe_bootstrap_theme:description_list' with {
+  title: title,
+  title_tag: title_tag|default('h2'),
```
donquixote: Do we need that default fallback here?
Looks like the sdc already takes care of it.

templates/patterns/description_list/pattern-description-list.html.twig
```diff
@@ -4,10 +4,10 @@
...
+  items: items,
+  orientation: orientation,
+  attributes: attributes,
+} only %}{% endembed %}
```
donquixote: If variables have the same name, we could just pass through like this:

```suggestion
{% embed 'oe_bootstrap_theme:description_list' %}{% endembed %}
```
For the embed/endembed, we always need closing tag? (I think yes)
And we cannot use include for sdc? (Looks like we can, looking at core)
```suggestion
{% include 'oe_bootstrap_theme:description_list' %}
```
If we need to override specific parameters we can still do that with `with`, if we omit the `only`.

components/description_list/description_list.component.yml
```diff
@@ -0,0 +1,30 @@
...
+      type: [string, array, object]
+      description: 'Title of the description list.'
+    title_tag:
+      title: 'Title tag'
+      type: string
+      description: 'The tag to use for the title. Defaults to h2.'
+    items:
+      title: 'List items'
+      type: array
+      description: 'Each item is composed by one or more terms and one or more definitions. Each term entry can have an icon.'
```
donquixote, line 17: I see more detailed items schema in the other PR https://github.com/openeuropa/oe_bootstrap_theme/pull/547/changes

donquixote: Maybe we can keep the pattern changes in a separate commit, and treat this PR as a continuation of the other PR.

components/description_list/description_list.story.twig
```diff
@@ -0,0 +1,42 @@
+{% embed 'oe_bootstrap_theme:description_list' with {
```
donquixote: Again, seems like `include` will be simpler than `embed`, because we are not overriding any blocks.

components/description_list/description_list.story.twig
```diff
@@ -0,0 +1,42 @@
...
+        {
+          label: 'Single label with icon',
+          icon: { name: 'geo-alt-fill' }
+        }
+      ],
+      definition: 'Description of first term text goes here.'
+    },
+    {
+      term: [
+        { label: 'First label without icon' },
```
donquixote, line 15: The new version of PatternDataMassager allows to inject markup with yaml taggedvalue like `!Markup ...`. We should use that to test if markup is preserved.

piotrsmykaj: I cannot find anything related to PatternDataMassager or !Markup. Could you send any code / documentation about it?

donquixote: Actually my comment might be out of place here.

The class is `PatternTestDataMassager` which was recently changed to support special values in yaml test data.
However, this here is a story written in twig, not yaml, and it is not meant for testing.
So my comment is indeed completely misplaced.

templates/patterns/description_list/pattern-description-list.html.twig
AaronGilMartinez: True

templates/patterns/description_list/pattern-description-list.html.twig
AaronGilMartinez: Seems duplicated code but this way we limit parameters to those indicated, no extra context for the component and we document what is expected. This is just the same convention we have been following so far, we can change it if we see a better approach.

components/description_list/description_list.component.yml
AaronGilMartinez: Done

components/description_list/description_list.story.twig
AaronGilMartinez: Done too.

templates/patterns/description_list/pattern-description-list.html.twig
donquixote: actually it is bcl taking care of it

donquixote:
test comment, please ignore

templates/patterns/description_list/pattern-description-list.html.twig
donquixote: Ok let's keep the explicit version.
But we should still use include over embed imo.

donquixote:
No need for `|default('h2')` in the sdc, bcl takes care of it.
(I cannot comment on that file for some reason)

AaronGilMartinez:
> No need for `|default('h2')` in the sdc, bcl takes care of it. (I cannot comment on that file for some reason)

Forgot about pattern.

components/description_list/description_list.twig
```diff
@@ -7,7 +7,7 @@
 {% include '@oe-bcl/description-list' with {
   title: title,
   title_link: title_link,
-  title_tag: title_tag|default('h2'),
+  title_tag: title_tag,
```
donquixote, line 10: No need, bcl takes care of h2 fallback.

templates/patterns/description_list/pattern-description-list.html.twig
```diff
@@ -4,10 +4,10 @@
...
  */
 #}
-{% include '@oe-bcl/description-list' with {
-  'title': title,
-  'title_tag': title_tag,
-  'items': items,
-  'variant': orientation,
-  'attributes': attributes,
-} only %}
+{% embed 'oe_bootstrap_theme:description_list' with {
```
donquixote: We should use include, not embed.

templates/patterns/description_list/pattern-description-list.html.twig
```diff
@@ -4,10 +4,10 @@
...
-  'items': items,
-  'variant': orientation,
-  'attributes': attributes,
+{% include 'oe_bootstrap_theme:description_list' with {
+  title: title,
+  title_tag: title_tag,
+  items: items,
+  orientation: orientation,
+  attributes: attributes,
 } only %}
```
donquixote, line 13: I just learned that:
```twig
{# keys can be omitted if it is the same as the variable name #}
{Paris}
```
https://twig.symfony.com/doc/3.x/templates.html#literals

But no need to update this PR, we can use this later if we want.

PR #551, opened 2026-04-10
OEL-4643: Media Copyright Accessibility improvements

piotrsmykaj:


docker-compose.yml
```diff
@@ -1,6 +1,6 @@
 services:
   web:
-    image: registry.fpfis.eu/digit/devsecops/httpd-php:8.3-dev
+    image: registry.dhs.tech.ec.europa.eu/digit/devsecops/httpd-php:8.3-dev
```
donquixote: Alternatively we could use `fpfis/httpd-php-ci:8.3` or `fpfis/httpd-php-dev:8.3`.

With `dev` we keep what we have now, but it means any CI that uses the image would need to replace `dev` with `ci` or suffer an efficiency penalty. This would be the case for #535 which uses a github action for visual regression testing.

With `ci` version, local developers need to replace the image with `docker-compose.override.yml` to be able to use xdebug.

templates/patterns/banner/pattern-banner.html.twig
```diff
@@ -29,6 +29,7 @@
   description: description,
   link: call_to_action,
   copyright: (variant == 'image' or variant == 'image_shade') ? copyright : '',
+  copyright_label: (variant == 'image' or variant == 'image_shade') ? 'Image credit:'|t : '',
```
donquixote, line 32: I don't know if we should hard-code this in every template, but I also can't think of a better solution right now.

Maybe keep it like this until we get a request to make it configurable.

piotrsmykaj (commented):
@donquixote I have replied to your comments.

docker-compose.yml
piotrsmykaj: https://citnet.tech.ec.europa.eu/CITnet/jira/browse/FPFISSUPP-34210?focusedId=8880239&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-8880239

The comment from Sebastien says:

> Yes, when the image is available under digit/devsecops/*, please use this one.
> 
> The fpfis/* repositories have been moved to digit/devsecops/* and are now archived/read-only as shown on the project page: https://gitlab.dhs.tech.ec.europa.eu/fpfis/httpd-php
> 
> So digit/devsecops/* should be used as the default choice.

I prefer it's paid by "CI efficiency" rather than the "Developers' Frustration" :-)

donquixote: Alright, thank you! Let's keep it.

donquixote (dismissed):
Will need rebase before merging.

donquixote (approved):

PR #554, opened 2026-04-21
OEL-4709: Added pagination v2 patterns.

tibi2303:
Jira issue(s):

### DIFF
--- a/templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
+++ b/templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
@@ -1,0 +1,46 @@
+{#
+/**
+ * @file
+ * Pagination V2 pattern.
+ */
+#}
+
+{% set _attributes = attributes|default(create_attribute()) %}
+{% if aria_label is not empty %}
+  {% set _attributes = _attributes.setAttribute('aria-label', aria_label) %}
+{% endif %}
+
+{% set _items = [] %}
+{% for item in items|default([]) %}
+  {% set _items = _items|merge([
+    item|merge({
+      'aria_label': item.aria_label|default(item.label|default('')),
+    }),
+  ]) %}
+{% endfor %}
+
+{% include '@oe-bcl/bcl-pagination-v2/pagination.html.twig' with {
+  'size': size,
+  'alignment': alignment,
+  'enable_prev_next_icon': enable_prev_next_icons,
+  'enable_first_last_icon': enable_first_last_icons,
+  'ellipsis_before': ellipsis_before,
+  'ellipsis_after': ellipsis_after,
+  'current_mobile_label': current_mobile_label,
+  'first': first ? first|merge({
+    'aria_label': first.aria_label|default('First'|t),
+  }),
+  'prev': prev ? prev|merge({
+    'aria_label': prev.aria_label|default('Previous'|t),
+  }),
+  'next': next ? next|merge({
+    'aria_label': next.aria_label|default('Next'|t),
+  }),
+  'last': last ? last|merge({
+    'aria_label': last.aria_label|default('Last'|t),
+  }),
+  'items': _items,
+  'attributes': _attributes,
+  'list_attributes': list_attributes,
+  'icon_path': bcl_icon_path,
+} only %}
--- a/templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
+++ b/templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
@@ -10,22 +10,36 @@
   {% set _attributes = _attributes.setAttribute('aria-label', aria_label) %}
 {% endif %}
 
+{% set _enable_prev_next_icon = enable_prev_next_icon ?? enable_prev_next_icons ?? true %}
+{% set _enable_first_last_icon = enable_first_last_icon ?? enable_first_last_icons ?? false %}
 {% set _items = [] %}
+{% if ellipsis_before %}
+  {% set _items = _items|merge([{
+    'item_type': 'ellipsis',
+    'aria_label': 'Ellipsis'|t,
+  }]) %}
+{% endif %}
 {% for item in items|default([]) %}
-  {% set _items = _items|merge([
-    item|merge({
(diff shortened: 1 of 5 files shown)


templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
```diff
@@ -0,0 +1,46 @@
...
+{% set _attributes = attributes|default(create_attribute()) %}
+{% if aria_label is not empty %}
+  {% set _attributes = _attributes.setAttribute('aria-label', aria_label) %}
+{% endif %}
+
+{% set _items = [] %}
+{% for item in items|default([]) %}
+  {% set _items = _items|merge([
+    item|merge({
+      'aria_label': item.aria_label|default(item.label|default('')),
```
AaronGilMartinez: I think is better to not set  second default, empty aria labels are worse than non-existent we can fallback to label only.

templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
```diff
@@ -0,0 +1,46 @@
...
+    }),
+  ]) %}
+{% endfor %}
+
+{% include '@oe-bcl/bcl-pagination-v2/pagination.html.twig' with {
+  'size': size,
+  'alignment': alignment,
+  'enable_prev_next_icon': enable_prev_next_icons,
+  'enable_first_last_icon': enable_first_last_icons,
```
AaronGilMartinez: It seems we have plural-singular missmatch. For me its singular.

tests/src/PatternAssertion/PaginationV2PatternAssert.php
```diff
@@ -0,0 +1,149 @@
...
+      'nav.pager.pagination-v2 > ul.pagination > li.page-item' => $expected_count,
+      'nav.pager.pagination-v2 > ul.pagination > li.page-item > .page-link:first-child' => $expected_count,
+    ], $crawler);
+
+    $items = $crawler->filter('nav.pager.pagination-v2 > ul.pagination > li.page-item');
+    $icon_pattern_assert = new IconPatternAssert();
+
+    foreach ($expected as $index => $expected_item) {
+      $item = $items->eq($index);
+      $this->assertCount(1, $item);
```
AaronGilMartinez: This is always going to be true since even if the index is out of bounds.

tests/src/PatternAssertion/PaginationV2PatternAssert.php
```diff
@@ -0,0 +1,149 @@
...
+
+      if (isset($expected_item['mobile_label'])) {
+        self::assertSame($expected_item['mobile_label'], trim($link->filter('.d-inline-block.d-md-none')->text()));
+      }
+
+      if (isset($expected_item['label']) && !isset($expected_item['desktop_label']) && !isset($expected_item['mobile_label'])) {
+        self::assertSame($expected_item['label'], trim($link->text()));
+      }
+
+      if (array_key_exists('aria_label', $expected_item)) {
```
AaronGilMartinez, line 108: Should we always check this to ensure accessibility? PaginationPatternAssert does.

tests/src/Kernel/fixtures/markup_rendering_patterns/pagination_v2.yml
```diff
@@ -0,0 +1,180 @@
+pagination_v2_default:
+  render:
+    '#type': pattern
+    '#id': pagination_v2
+    '#fields':
+      aria_label: 'Search results pages'
+      enable_prev_next_icons: false
+      enable_first_last_icons: false
```
AaronGilMartinez: Better to set it to true, otherwise it will fallback to BCL.

AaronGilMartinez (changes requested):

templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
tibi2303: done.

templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
tibi2303: done. added also the use cases for ellipsis and next prev

tests/src/PatternAssertion/PaginationV2PatternAssert.php
tibi2303: done.

tests/src/PatternAssertion/PaginationV2PatternAssert.php
tibi2303: indeed, added.

tests/src/Kernel/fixtures/markup_rendering_patterns/pagination_v2.yml
tibi2303: set it to true.

templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
```diff
@@ -0,0 +1,60 @@
...
+ */
+#}
+
+{% set _attributes = attributes|default(create_attribute()) %}
+{% if aria_label is not empty %}
+  {% set _attributes = _attributes.setAttribute('aria-label', aria_label) %}
+{% endif %}
+
+{% set _enable_prev_next_icon = enable_prev_next_icon ?? enable_prev_next_icons ?? true %}
+{% set _enable_first_last_icon = enable_first_last_icon ?? enable_first_last_icons ?? false %}
```
AaronGilMartinez: No need to add BC it's different version,

AaronGilMartinez (changes requested):

templates/patterns/pagination_v2/pattern-pagination-v2.html.twig
tibi2303: true sorry..

AaronGilMartinez: nothing to be sorry

AaronGilMartinez (approved):

PR #560, opened 2026-05-06
OEL-4732: Remove selenium version.

tibi2303:
Jira issue(s):

### DIFF
--- a/runner.yml.dist
+++ b/runner.yml.dist
@@ -91,4 +91,3 @@ toolkit:
     BEHAT: false
     PHPUNIT: true
     SELENIUM: true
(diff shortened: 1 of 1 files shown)


runner.yml.dist
```diff
@@ -91,4 +91,3 @@ toolkit:
     BEHAT: false
     PHPUNIT: true
     SELENIUM: true
-    SELENIUM_VERSION: 4.1.3-20220405
```
donquixote, line 94: Seems that version we remove here is still the same that toolkit uses in current version.
But that's ok.

donquixote (approved):

PR #565, opened 2026-05-13
OEL-4746: Added test for logos path exists.

tibi2303:
Jira issue(s):

### DIFF
--- a/tests/src/Unit/ThemeAssetsTest.php
+++ b/tests/src/Unit/ThemeAssetsTest.php
@@ -17,10 +17,16 @@ class ThemeAssetsTest extends UnitTestCase {
   public function testCanaryAssetFilesExist(): void {
     $theme_root = dirname(__DIR__, 3);
     $expected_files = [
+      'assets/bcl/bcl-button/bcl-button.html.twig',
+      'assets/css/oe_bootstrap_theme.style.min.css',
       'assets/logos/ec/logo-ec--en.svg',
       'assets/logos/ec/logo-ec--fr.svg',
       'assets/logos/eu/logo-eu--en.svg',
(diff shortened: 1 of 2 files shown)


tests/src/Unit/ThemeAssetsTest.php
```diff
@@ -0,0 +1,34 @@
...
+  /**
+   * Tests that key generated asset files exist.
+   */
+  public function testCanaryAssetFilesExist(): void {
+    $theme_root = dirname(__DIR__, 3);
+    $expected_files = [
+      'assets/logos/ec/logo-ec--en.svg',
+      'assets/logos/ec/logo-ec--fr.svg',
+      'assets/logos/eu/logo-eu--en.svg',
+      'assets/logos/eu/mobile/logo-eu--en.svg',
```
donquixote, line 25: Seems ok to me.
We can always add more files to check, the list does not need to be "complete" in this PR.

However, I would at least have one file outside of these logos, to distinguish "BCL was not added at all" vs "BCL is there but logos are missing.".

Maybe add one BCL template in assets, and one generated css file in assets.

Should we check anything in node_modules or do we not care? Could there be a scenario when we run this test but node_modules is expected to be empty?

donquixote: For the existing canary files, they are ok but probably second and third are redundant. If first is present, probably the other too as well. The fourth maybe not because it's in a subdir so a bit special perhaps.

But it's ok to keep them all.

The main reason why we don't put _all_ files here is that we don't want to update the test if BCL intentionally removes or adds one of the files.

donquixote (approved):

PR #571, opened 2026-06-12
OEL-4815: Port alert pattern to SDC.

tibi2303:
Jira issue(s):
- [D8AGE-???](https://webgate.ec.europa.eu/CITnet/jira/browse/D8AGE-???)

PR #572, opened 2026-06-12
OEL-4816: Port badge pattern to SDC.

tibi2303:
Jira issue(s):


piotrsmykaj:
hi @tibi2303 
The BCL badge template accepts the parameter "assistive_text", but it's not exposed as a prop and is not passed in badge.twig. It was the same scenario in the old pattern, but it's a good approach to improve it and add the parameter.

components/badge/badge.component.yml
```diff
@@ -0,0 +1,60 @@
...
+      'meta:enum':
+        primary: 'primary (default)'
+        secondary: secondary
+        success: success
+        danger: danger
+        warning: warning
+        info: info
+        light: light
+        dark: dark
+        link: link
```
piotrsmykaj: alert.component.yml in #571 has the same enum but capitalised:
```
     'meta:enum':
        primary: 'Primary (default)'
        secondary: Secondary
        success: Success
        danger: Danger
        warning: Warning
        info: Info
        light: Light
        dark: Dark
        link: Link
``` 
Minor inconsistency between the two SDCs, but it would be good to make them consistent.

components/badge/badge.story.twig
```diff
@@ -0,0 +1,9 @@
+{% include 'oe_bootstrap_theme:badge' with {
```
piotrsmykaj: We could present more variants of the component in badge.story.twig:

Image

What do you think?

tests/src/Kernel/fixtures/markup_rendering_components/badge.yml
```diff
@@ -0,0 +1,100 @@
...
+      class: Drupal\Tests\oe_bootstrap_theme\PatternAssertion\BadgePatternAssert
+      expected:
+        settings:
+          background: danger
+          rounded_pill: true
+          dismissible: true
+          outline: true
+        label: 'A simple badge!'
+        url: 'https://example.com'
+        title: 'The badge title'
```
piotrsmykaj:
```suggestion
        background: danger
        rounded_pill: true
        dismissible: true
        outline: true
        label: 'A simple badge!'
        url: 'https://example.com'
        title: 'The badge title'
```
SDC doesn't support "settings" and "fields" as they were replaced by "props" and "slots", so calling it "settings" here is a bit confusing as it's linked to the old patterns. I suggest changing it to the flat array and creating `tests/src/ComponentAssertion/BadgeComponentAssert.php` 

I checked that in the description_list.yml I called `PatternAssertion\DescriptionListAssert`:
```
  assertions:
    component:
      class: Drupal\Tests\oe_bootstrap_theme\PatternAssertion\DescriptionListAssert
      expected:
        items:
          - term:
              - label: 'Single label with icon'
                icon: 'geo-alt-fill'
            definition: 'Description of first term text goes here.'
          [...]
```
but here we only test the "items" field, which belongs to "fields" and is directly under "expected" in the pattern test, so naming is fine in this case also for the component, but an alternative option is to create a separate `ComponentAssertion\DescriptionListAssert` for the component to make it consistent for all components. What do you think?

piotrsmykaj (approved):

PR #576, opened 2026-06-16
OEL-4814: Port Accordion pattern to SDC

piotrsmykaj:


tests/src/Kernel/fixtures/markup_rendering_components/accordion.yml
```diff
@@ -0,0 +1,139 @@
...
+      '.bcl-heading': 1
+      'h2.bcl-heading': 1
+      'div.accordion.accordion-flush': 0
+      'button[data-action="expand"]': 1
+      'button[data-action="collapse"]': 1
+      'div.accordion-item': 3
+      'div.accordion-item h2.accordion-header': 3
+      'div.accordion-item button.accordion-button': 3
+      'div.accordion-item div.accordion-collapse.collapse': 3
+      'div.accordion-item div.accordion-collapse.collapse.show': 1
```
tibi2303, line 133: small stuff but can we add also some selectors that check data-target and the ids are the same

'button[data-action="collapse"][data-target="accordion-test-accordion"]': 1

'div#accordion-test-accordion.accordion': 1

tibi2303 (approved):

